Issue #74, opened 2018-06-26
Consider switching from Jupyter Notebook to JupityerLab


joernweissenborn:
Maybe write some plugins to connect with the JupyterLab Data Model.

jsnel:
The use of Jupyter Notebooks is currently very much supportive, they are not yet an integral part of the package, so at this stage I see no reason not to switch to JupityerLab, which is clearly the future of Python.
Ref: https://github.com/jupyterlab/jupyterlab 
"The beta releases are suitable for general usage. For JupyterLab extension developers, the extension APIs will continue to evolve until the 1.0 release." As we are not going to develop extensions (AFAIK) we should be fine to use JupityerLab for the purpose of quickstart files, tutorials and documentation. 

I have considered this issue and am signing off, assigning to @joernweissenborn and @s-weigand. When in agreement, close this issue and create issues for the actual conversion of notebooks and/or creation of JupyterLab notebooks from scratch.

joernweissenborn:
I haven't evaluated it yet, but what we in ASAP is a conviniend way to feed data into Glotaran. I hope that Jupyter Lab can help us with this.

Issue #73, opened 2018-06-26
Update Vagrant file


joernweissenborn:
The user experience can be much better.

PR #76, opened 2018-06-28
make using vagrant and jupyterlab easy

s-weigand:
This PR concerns the issues #74 and  #73.

I updated the Vagrantfile to use ubuntu 18.04 LTS and use jupyterlab instead of notebooks.
I also removed some redundant package installations in ubuntu-bootstrap.sh and moved the installation of lmfit-varpro to setup.py (one step closer to a PYPI package).

For jupyterlab on vagrant not to conflict with possible running jupyter instances of users with default config, I can changed port 8888 to 9999 (over 9000 :P ) and removed the authetication token.
Thus a user now can just run "vagrant up" and click on "open_vagrant-jupyter.html" to get to jupyterlab or get a list with possible reasons why he/she isn't redirected.
Feel free to pretty up "open_vagrant-jupyter.html" with some css.

### DIFF
--- a/.jupyter_config.py
+++ b/.jupyter_config.py
@@ -1,0 +1,19 @@
+# Password to use for web authentication
+# the password can be generated with:
+# >>>from notebook.auth import passwd
+# >>>passwd("glotaran")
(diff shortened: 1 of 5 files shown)


joernweissenborn:
Great work, thank you alot!

PR #77, opened 2018-06-30
Fix lmfit varpro installation

s-weigand:
Hi guys,

i found that the installation of lmfit-varpro from my PR #76  didn't work, but now it does ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ In folder, run:
 
 ## Vagrant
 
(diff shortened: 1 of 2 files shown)


joernweissenborn:
I don't see what the actual problem was, but the changeset seams ok to me, so I will merge.

s-weigand:
The problem was that `lmfit-varpro`, didn't get installed.
The installation didin't throw an error, but when running `fit_data_in_vivo_cyano_data.ipynb` it complained that `lmfit-varpro` wasn't installed.

Issue #139, opened 2018-08-10
Remove the git installation of lmfit-varpro in tests


s-weigand:
### Description
Since ``lmfit-varpro`` is on PyPi now there is now need manually install the git version of it.
Also this could lead to falsly passing test, if the minimum required version of ``lmfit-varpro`` wasn't bumped in ``setup.py``.

joernweissenborn:
`lmfit-varpro` dependency is fine.

https://github.com/glotaran/glotaran/blob/master/setup.py#L83

PR #140, opened 2018-08-10
Generate correct coverage

### LINKED ISSUE #138: Fix the generated coverage reports
### Description
As of now the produced coverage reports can't find the run code, which is why the coverage is so low.

s-weigand:
* Excluded Tests themself from coverage
* Fixed generation of coverage with Tox
* Fixed generation of coverage with CIs
* Removed lmfit-varpro force installation

**Closing issues**

closes #138 #139

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ branch = True
 include = glotaran/*
 omit = setup.py
        */tests/*
+       */test/*
 
 [report]
 # Regexes for lines to exclude from consideration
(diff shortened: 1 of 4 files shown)


.coveragerc
```diff
@@ -3,6 +3,7 @@ branch = True
 include = glotaran/*
 omit = setup.py
        */tests/*
```
joernweissenborn: `tests` doesn't exist anymore, so we should not omit it in case we move stuff there later.

s-weigand: ``tests`` does exists at ``glotaran/models/spectral_temporal/tests``

joernweissenborn: ok, i got it wrong

Issue #144, opened 2018-08-11
Support reading sdt as TCSPC data


jsnel:
The data generated by [Becker-Hickl TCSPC setups ](https://www.becker-hickl.com/products.htm) is a binary file with extension *.sdt. Usually the binary data can be exported to ascii files (with extensions e.g. txt, dat, ascii, kin). 

In the zip attached are some sdt files and their ascii counterpart:
[sdt_as_TCPSC_data_example.zip](https://github.com/glotaran/glotaran/files/2280569/sdt_as_TCPSC_data_example.zip)
In this archive you will find four files: 
kinetic trace: **368 ex 640 em 700.sdt** in ascii: **680_700.kin**
corresponding IRF: **IRF EX640.sdt** in ascii: **IRF640.kin**

Issue #143, opened 2018-08-11
Support for Becker-Hickl .sdt file format 


jsnel:
### Background info
The *.sdt Becker-Hickl file format is quite common for time-correlated single photon counting ([TCSPC](https://en.wikipedia.org/wiki/Fluorescence-lifetime_imaging_microscopy#TCSPC)) setups. It is used both in a microscopy settings, in fluorescence lifetime imaging microscopy (FLIM) data aqusition, as well as measuring regular time-gated fluorescence decays (at multiple wavelengths using a monochromator). Support for the proprietary sdt file format would need to cover both use cases, so a collection of issues is identified.

The company has a very extensive Handbook online in which most information can be found:
https://www.becker-hickl.com/handbookphp.htm   

### Dependent issues / tasks
Reading raw data:
- [ ] #124 get a reader for *.sdt files
Reading FLIM data:
- [ ] #142  Support for FLIM ascii data format 
- [ ] # Support reading sdt as FLIM data
Reading TCSPC data:
- [ ] #144 Support reading sdt as TCSPC data

Issue #142, opened 2018-08-11
Support for FLIM ascii data format


jsnel:
### FLIM Image format 
Given a FLIM image, it is often desirable to select those pixels associated with the subject of interest for modeling. For example, given a FLIM image of a cell, only those pixels interior to the cell wall are typically representative of the fluorescence decay of interest, and accordingly only these pixels are usually selected for modeling. Pre-processing dedicated to pixel selection is performed prior to exporting the data to a FLIM Image format. The indices of selected pixels are then included in the ascii file format.

The format of the input ﬁle is as follows.
```
line 1: reserved for comments, not read
line 2: reserved for comments, not read
line 3: the character string “FLIM Image”
line 4: dimension of image as x y (space-delimited)
line 5: numb er of timep oints t in image
line 6: numb er of pixels x selected for analysis
line 7: vector of the times at which measurements were made
line 8: index of selected pixel and decay trace from this pixel Ψ(,p)
... ...
Line (8+numb er of selected pixels): the character string “Intensity map” t1, t2,..., tm Note that each column represents a decay trace, and each row represents a time-gated spectrum. 
```

s-weigand:
The `Intensity map`, as I just found out, is a map of the pixels (image) with the sum over the time trace  as value for each pixel. So it's basically the timesum of the FLIM.

Btw how exactly did the selection process work?
Fothe the given data I found:
`mean(Intensity map)` = 245.5
`median(Intensity map)`= 73.0
`min(selected pixels)` = 798

PS.: If you want i can drop my data exploration notebook here.

SerLap:
yes the intensity map show an time integral, but can be changed in the gui to have max number in the trace per pixel. 
Selection of the pixels is on the basis  of the intensity image just selecting all pixels that has higher than thresh hold value in the intensity matrix.

s-weigand:
With PR #151 we have legacy support 😄

Issue #64, opened 2017-06-26
Pin down version number of package dependencies


jsnel:
In setup.py depend on package x.y.z

We should not allow future versions in case one of our dependencies breaks their API.

joernweissenborn:
will be fixed with #71

PR #151, opened 2018-09-02
Add sdt support

### LINKED ISSUE #124: get a reader for *.sdt files
Not to reinvent the wheel we should use work already done by other people to read binany files.
For ``*.sdt`` files that would be [Christoph Gohlke](cgohlke@uci.edu) who wrote a reader for [``*.sdt`` ](https://www.lfd.uci.edu/~gohlke/code/sdtfile.py.html) and also some other [fileformats](https://www.lfd.uci.edu/~gohlke/). His files are under BSD licence, so it shouldn't be a problem to use them.

@jsnel since you are the projects manager, can you ask him, if he is fine with us using his work, when we reference him in credits? I think that is a nicer way than just using it.

s-weigand:
* added custom base class for high dimentional data `glotaran.model.dataset.HighDimentionalDataset`
* added custom class for FLIM data `glotaran.model.specialized_datasets.FLIMDataset`
* implemented `pd.DataFrame` as intermediate repressentation for Datasets, IMO this will make it easier for to implement readers for custom formats in the future
* in `glotaran.dataio.file_readers`
  - added `df_to_SpectralTemporalDataset`
  - added `df_to_FLIMDataset`
  - added `sdt_to_df`
  - added `read_sdt`
* added support for FLIM legacy data (glotaran 1.5), `glotaran.dataio.legacy_readers.FLIM_legacy_to_df`
* added mapper for FLIM data to pixel position (tuple) `glotaran.dataio.mapper.get_pixel_map`
* pinned down versions of setup - and install requirements in setup.py

**Testing**

* added 194 tests for the readers (droping some of the `spectral_unit` and `time_unit` will greatly reduce the amount in tests and that for make the test suite pass faster, but that open for discussion)
* didn't add tests for the new Datasets since they will be totally rewritten by PR #150 anyway

**Closing issues**

closes #124 #142 #143 #144 #64


jsnel:
Please rename glotaran.model.dataset.HighDimentionalDataset to glotaran.model.dataset.HighDimensionalDataset

glotaran/dataio/file_readers.py
```diff
@@ -0,0 +1,378 @@
...
+import numpy as np
+import pandas as pd
+from .external_file_readers.sdt_reader import SdtFile
+from .mapper import get_pixel_map
+from ..models.spectral_temporal.spectral_temporal_dataset import SpectralTemporalDataset
+from ..model.specialized_datasets import FLIMDataset
+from ..model.dataset import DimensionalityError
+
+
+def df_to_SpectralTemporalDataset(input_df: pd.DataFrame, dataset_label: str,
```
joernweissenborn: In glotaran we prefer verbose names. please change to `dataframe_to_...`

glotaran/model/dataset.py
```diff
@@ -49,96 +58,198 @@ def get_calculated_axis(self):
         """Get the axis along the estimated matrices are grouped"""
         raise NotImplementedError
 
-    def get_axis(self, label):
+    def get_axis(self, axis_label: str) -> Union[list, np.ndarray]:
```
joernweissenborn: the axis MUST be ndarrays

glotaran/model/dataset.py
```diff
@@ -1,42 +1,51 @@
...
 
 
+class DimensionalityError(Exception):
+    """
+    Custom exception if data have the wrong dimensionality
+    """
+    pass
+
+
 class Dataset(object):
```
joernweissenborn: This class have undergone alot of changes. you will have to adapt.

glotaran/model/dataset.py
```diff
@@ -49,96 +58,198 @@ def get_calculated_axis(self):
...
+
+            rsv: np.ndarray
                 right singular values
 
         """
         lsv, svals, rsv = np.linalg.svd(self.get().T)
         return lsv, svals, rsv.T
 
 
+class HighDimensionalDataset(Dataset):
```
joernweissenborn: This is a helper class and does not concern models. Move to somwhere in `dataio`

glotaran/model/specialized_datasets.py
```diff
@@ -0,0 +1,123 @@
+"""
```
joernweissenborn: This also belongs to dataio.

glotaran/model/dataset.py
s-weigand: Guess by that you mean a rebase after PR #150 was merged?

s-weigand:
* moved docs linkcheck to a different test and made it `allowed to fail` in travis, since it depends on eternal resources

joernweissenborn:
Agreed

Joris Snellenburg <notifications@github.com> schrieb am Mo., 10. Sep. 2018,
09:25:

glotaran/model/dataset.py
s-weigand: another idea would be to make `get_axis` and `set_axis` privat ( `_get_axis` / `_set_axis` ) and let the user only see the custom implementation, as is done in `SpectralTemporalDataset` with `time_axis` and `spectral_axis`.

joernweissenborn: I am generally not caring so much. I have set it now to ndarray, since I use excplicitly things like `axe.size`. The thing is, that an axe in our context must have a size which can be retrieved by a standard method somehow and it must contain hashable items (actually only for the calculated axis ('time' in terms of a kinetic model).

Leaving that aside, `Dataset` is an abstract class without ABC. It provides a standart implementation for convinience (escpecially simulations). I don't want to strictly split it into interface and implementation, but maybe I should. In that case, I would leave the setter for axies out and only put the getter into. I experimentated with abstract properties, that would be a nice way to go.

As for this PR. consider Dataset as an interface like that:

```python
class Dataset:
    get_axis(label: str) -> ndarray
    data() -> ndarray
```

s-weigand:
* rebased to the current version (just after PR #153 merge)
* moved all datasets to a package `glotaran.datasets` (might need some discussion about that, since now the datasets and their discriptor are separated)
* removed `tox` from `.travis.yml`
* the 'docs-fix' is just a hot fix, for the docs to generate them at all (the underlying problem is that some imports fail in `glotaran.analysis` and the new version of sphinx does't like the hack of @joernweissenborn to use the string version of the importpath )

PR #216, opened 2019-10-24
fixed wrong encrypted PyPi password

s-weigand:
Sorry guys the PyPi password I reencrypted was wrong due to none escaped characters.
Note to myself always use single quotes >.<

Anyway now it should work, just try it after the merge.
```bash
$ bumpversion patch
```


joernweissenborn:
fck win

PR #248, opened 2019-11-14
Small fix for gh actions

s-weigand:
Give a brief summary of the changes.
- Rename gh actions from `Python package` to `Tests`
- Make the also trigger at PRs not only pushes


s-weigand:
Wrong branch, should be `fix_performance`

PR #261, opened 2019-11-19
Basic implementation of Irf Convolution

joernweissenborn:
* created a method to load extra datasets for the optimization
* implemented `conv1` and `conv2` from TIMP
* Added `method` property to `IrfMeasured`
* Added CLI option to declare extra data from commandline


joernweissenborn:
So here is basically how I want to implement measured IRFs it will be easy to extend so one can create a 'convolution addon' to support more convolutions.

@jsnel and @SerLap please review this part:

https://github.com/glotaran/pyglotaran/pull/261/files#diff-eb5e9d7255c7060f2a32f2e78a925ffeR24-R70

jsnel:
Regarding the IrfMultiGaussian implementation. 

The way that it is documented in TIMP (and also implemented in TIM) is as follows:

A (simple) gaussian IRF is specified by two numeric values: location and width (FWHM).
- number of parameters needed: 2

If the IRF is specified as a *double* gaussian four numeric values are needed to specify the IRF, corresponding to the location (mean) of the IRF, the width (FWHM) of the first Gaussian, the width of the second Gaussian, and the relative amplitude of the second Gaussian, respectively. A *triple* gaussian adds another width and another relative amplitude parameter, etc.
- number of parameters needed: 2+2, 2+4

In case of *multiple* gaussian the IRF is specify by two numeric values for the first IRF (location (mean) and FWHM (width) and three (3) subsequent values for each additional gaussian modeled, corresponding to the relative scaling to the first gaussian, the shift (in time) relative to the first gaussian and the FWHM of the additonal Gaussian, respectively
- number of parameters needed: 2+3, 2+6

Our current implementation 'breaks' with this in the sense that instead of a shift parameter the direct location is specified instead (which is fine, as long as it is clearly documented that it is a absolute parameters rather than a relative shift with respect to the main IRF position). Using parameter relations it may be possible for a user to restore the old behavior (irf position2 = irf position1 + shift2)

It needs further investigation whether the scale parameter is implemented as expected. The scale parameter is corrected use in calculation; scale (scal) is used there to multiply each additional gaussian prior to adding it to the total vector. Reference:
https://github.com/glotaran/TIMP/blob/master/pkg/R/calcCirf.R
However, breaking with TIMP tradition the scale of the first IRF (in TIMP always 1) also needs to be specified, requiring an additional (by convention always fixed) parameter. This is fine as long as it is well documented that to achieve the original behavior all scale parameters need to be specified whereby the first one is fixed to 1.

glotaran/analysis/matrix_calculation.py
```diff
@@ -129,9 +132,8 @@ def _calculate_matrix(matrix_function, dataset_descriptor, axis, extra, index=No
     args = {
         'dataset_descriptor': dataset_descriptor,
         'axis': axis,
+        'extra': extra,
     }
-    for k, v in extra:
```
jsnel: What was the prior functionality you removed here?
It seems 'extra' was already part of the _calculate_matrix function args prior to this change.

glotaran/builtin/models/kinetic_image/kinetic_image_matrix.py
```diff
@@ -3,17 +3,18 @@
 import numpy as np
 import numba as nb
 
-from .irf import IrfMultiGaussian, IrfGaussianCoherentArtifact
+from .irf_gaussian import IrfMultiGaussian
+from .irf_measured import IrfMeasured
 
 
 def kinetic_image_matrix(
-        dataset_descriptor=None, axis=None, index=None, irf=None):
```
jsnel: Strange substitution ... how is 'irf' equivalent to 'extra'?

glotaran/builtin/models/kinetic_image/kinetic_image_matrix.py
```diff
@@ -67,21 +68,11 @@ def kinetic_matrix(
             compartments.append(baseline_compartment)
             matrix = np.concatenate((matrix, baseline), axis=1)
 
-    if isinstance(dataset_descriptor.irf, IrfGaussianCoherentArtifact):
```
jsnel: I assume this is further cleanup resulting from the move of coherent artifact from kinetic_image to kinetic_spectrum class?

glotaran/builtin/models/kinetic_image/test/test_irf.py
```diff
@@ -0,0 +1,80 @@
...
+            },
+        },
+        'megacomplex': {
+            'mc1': {'k_matrix': ['k1']},
+        },
+        'k_matrix': {
+            "k1": {'matrix': {("s1", "s1"): '1', }}
+        },
+        'irf': {
+            'irf1': {'type': 'measured', 'method': 'conv1'},
```
jsnel: should conv2 not also be unit tested?

glotaran/builtin/models/kinetic_image/test/test_irf.py
```diff
@@ -0,0 +1,80 @@
...
+    assert model.valid(wanted)
+
+    print(model.validate(initial))
+    assert model.valid(initial)
+
+    print(model.markdown(initial))
+
+    extra = {'irf1': irf}
+
+    dataset = model.simulate('dataset1', wanted, axis, clp, extra=extra)
```
jsnel: Could you also test a normal parameterized gaussian irf, but then fitted with a measured irf? 
This seems to only simulate with a measured_irf and then also fit with measured_irf.

glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_matrix.py
```diff
@@ -8,9 +8,9 @@
 from .spectral_irf import IrfGaussianCoherentArtifact
 
 
-def kinetic_spectrum_matrix(dataset_descriptor=None, axis=None, index=None, irf=None):
+def kinetic_spectrum_matrix(dataset_descriptor=None, axis=None, index=None, extra=None):
 
-    clp_label, matrix = kinetic_image_matrix(dataset_descriptor, axis, index, irf)
```
jsnel: what was irf used for before, because irf was inside dataset_descriptor right?

glotaran/io/reader.py
```diff
@@ -36,4 +36,4 @@ def decorator(reader):
     return decorator
 
 
-file_reader(extension='nc', name='netCDF4')(xr.open_dataset)
```
jsnel: Why was this change needed? What's the difference? Is this only applicable to the extra file?
"Open an DataArray from a file or file-like object containing a single data variable.
This is designed to read netCDF files with only one data variable. If multiple variables are present then a ValueError is raised."
Ref:
http://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html
http://xarray.pydata.org/en/stable/generated/xarray.open_dataarray.html

glotaran/analysis/test/test_scheme.py
```diff
@@ -23,13 +23,18 @@ def scheme(tmpdir_factory):
     dataset_path = path.join('dataset.nc')
     xr.DataArray([1, 2, 3]).to_dataset(name='data').to_netcdf(dataset_path)
 
+    extra_path = path.join('extradata.nc')
```
jsnel: Is this the only place where a nc datafile is generated and tested? This is a simple single data variable format of the nc format, not an entire full dataset. Is the latter also tested otherwise it may not be tested if open_dataarray vs open_dataset creates a problem down the line.

glotaran/builtin/file_formats/sdt/test/test_file_readers.py
```diff
@@ -17,8 +17,8 @@ def test_read_sdt(test_file_path, result_file_path, index):
                             dtype={"Delay": np.float, "Data": np.uint16})
     result_df.Delay = result_df.Delay * 1e-9
     result_traces = pd.DataFrame([result_df.Data.values], columns=result_df.Delay)
-    assert isinstance(test_dataset, xr.Dataset)
```
jsnel: I am unsure if xr.Dataset can be replaced with cr.DataArray without unintended consequences

glotaran/analysis/matrix_calculation.py
joernweissenborn: I had a different scheme in mind with `extra` originally, but I altered it. Prey they I don't alter it further... )

glotaran/analysis/optimize.py
```diff
@@ -38,6 +38,8 @@ def optimize(scheme, verbose=True, client=None):
 
 def optimize_task(initial_parameter, scheme, verbose):
 
+    scheme.prepare_data()
```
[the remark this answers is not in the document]
joernweissenborn: Yes, because I want to have the dataset svd as part of the result object. The copy is in principle unnecessary. Actually under the hood I think xarray already prevents copying here.

glotaran/analysis/scheme.py
```diff
@@ -85,7 +87,22 @@ def from_yml_file(cls, filename: str) -> 'Scheme':
...
         group_tolerance = scheme.get('group_tolerance', 0.0)
-        return cls(model=model, parameter=parameter, data=data,
+
+        extra = scheme.get('extra', {})
+
+        for label, path in extra.items():
+            path = pathlib.Path(path)
+
+            fmt = path.suffix[1:] if path.suffix != '' else 'nc'
+            if 'data_format' in scheme:
```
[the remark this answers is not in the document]
joernweissenborn: The thing is: if you define `data_format` in the scheme then ALL data is assumed to be in that format, otherwise it will deduced from the suffix (defaulting to `nc`). So having mixed data formats is possible, if the paths have the right suffix.

glotaran/analysis/matrix_calculation.py
```diff
@@ -25,7 +26,7 @@ def calculate_index_independend_ungrouped_matrices(scheme, parameter):
             model.matrix,
             descriptor,
             axis,
-            {},
```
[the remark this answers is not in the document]
joernweissenborn: Yes

glotaran/analysis/scheme.py
```diff
@@ -24,6 +24,7 @@ def __init__(self,
                  model: 'glotaran.model.Model' = None,
                  parameter: ParameterGroup = None,
                  data: typing.Dict[str, typing.Union[xr.DataArray, xr.Dataset]] = None,
+                 extra: typing.Dict[str, xr.DataArray] = None,
```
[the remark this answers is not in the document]
joernweissenborn: A `file_reader` MUST return a `DataArray`. The requirement for extra is that it can be loaded from disk with a `file_reader`. I could rename it to `extra_data`. Everything else is decided by the model, I don't care what they represent.

glotaran/analysis/scheme.py
```diff
@@ -115,6 +132,14 @@ def data(self) -> typing.Dict[str, typing.Union[xr.DataArray, xr.Dataset]]:
     def data(self, data: typing.Dict[str, typing.Union[xr.DataArray, xr.Dataset]]):
         self._data = data
 
+    @property
+    def extra(self) -> typing.Dict[str, xr.Dataset]:
```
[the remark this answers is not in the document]
joernweissenborn: That needs to be changed.

glotaran/builtin/models/kinetic_image/test/test_irf.py
joernweissenborn: This test is only basic and to be extended after we accepted the convolution functions.

glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_matrix.py
joernweissenborn: This as just a placeholder and never used.

glotaran/io/reader.py
joernweissenborn: Readers must return arrays now, because not everything they read are datasets for fitting. There must be some restrains on the dataset, if people really want to read from multivalue sets and are not able to extract the data to fit as array we can think about supporting datasets at this time.

glotaran/builtin/file_formats/sdt/test/test_file_readers.py
joernweissenborn: It can, you can always extract the things you want to fit as array with your favorite tool. The sdt file reader will suffer a little from that but nothing to busy. In the end, outside of `nc` a file represents 1 array, so the whole file abstraction is now centered around that a file represents one array.

glotaran/analysis/optimize.py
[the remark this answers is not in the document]
jsnel: The way I see it, as long as you ensure that your dataset contains the SVD object prior to analysis you can prevent the additional SVD computation during the call to optimize. Only with very large datasets (e.g. FLIM or stack of temporal scans) would doing the SVD computation at all become a bit of an annoying. Probably in >95% use cases it is a non issue, but perhaps we should make it optional at some point.

glotaran/analysis/scheme.py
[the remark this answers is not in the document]
jsnel: Ok, reminder to make this change within this PR.

glotaran/analysis/test/test_scheme.py
jsnel: I created https://github.com/glotaran/pyglotaran/issues/265 to address this

glotaran/analysis/scheme.py
```diff
@@ -191,8 +223,12 @@ def prepared_data(self) -> typing.Dict[str, xr.Dataset]:
             new_dims += [dim for dim in dataset.dims
                          if dim != self.model.matrix_dimension
                          and dim != self.model.global_dimension]
-            data[label] = dataset.transpose(*new_dims)
-        return data
+            if copy:
+                data[label] = dataset.transpose(*new_dims)
+            else:
+                self.data[label] = dataset.transpose(*new_dims)
```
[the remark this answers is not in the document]
joernweissenborn: to make sure all np.arrays are 'ivostyle'

glotaran/builtin/models/kinetic_image/irf_gaussian.py
```diff
@@ -0,0 +1,88 @@
...
+                raise Exception(f'len(centers) ({len_centers}) not equal '
+                                f'len(widths) ({len_widths}) none of is 1.')
+            if len_centers == 1:
+                centers = [centers[0] for _ in range(len_widths)]
+                len_centers = len_widths
+            else:
+                widths = [widths[0] for _ in range(len_centers)]
+                len_widths = len_centers
+
+        scale = self.scale if self.scale is not None else [1.0 for _ in centers]
```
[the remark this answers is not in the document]
joernweissenborn: yes, this a TODO

glotaran/builtin/models/kinetic_image/irf_gaussian.py
```diff
@@ -0,0 +1,88 @@
...
+from glotaran.model import model_attribute
+from glotaran.parameter import Parameter
+
+
+@model_attribute(properties={
+    'center': List[Parameter],
+    'width': List[Parameter],
+    'scale': {'type': List[Parameter], 'allow_none': True},
+    'normalize': {'type': bool, 'default': False},
+    'backsweep': {'type': bool, 'default': False},
```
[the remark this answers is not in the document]
joernweissenborn: You introduced it :). We need a way ofdocumenting this autogerated stuff, thiss is left for another pr...

glotaran/builtin/models/kinetic_image/test/test_irf.py
joernweissenborn: yes

jsnel:
We're seeing numba related errors under Python 3.8. To reproduce locally: 

# conda activate python_3.8_env
git checkout -b irf_convolution_1 origin/irf_convolution_1
pytest glotaran/builtin/models/kinetic_image/test/test_irf.py
# observe errors

Please have a look @joernweissenborn

joernweissenborn:
@SerLap and @jsnel:

It is time now for benchmarking. I need a test dataset plus fitting goals, for both `conv1` and `conv2`.

jsnel:
@joernweissenborn please check how you wish to salvage these changes in light of the refactored main branch. 
Certain elements, such as the convolution code may be recycled.

PR #310, opened 2020-02-09
Fix kinetic spectrum matrix irf

jsnel:
Small bugfix 

irf was none when running test_spectral_irf.py ... the test would pass but the entire kinetic trace would remain zero, this fixes that and tests still pass


jsnel:
Closed in favor of https://github.com/glotaran/pyglotaran/pull/313

PR #313, opened 2020-02-12
Fix kinetic spectrum matrix irf2

jsnel:
Small bugfix

irf was none when running test_spectral_irf.py ... the test would pass but the entire kinetic trace would remain zero, this fixes that and tests still pass

cherry-picked commits from fix_kinetic_spectrum_matrix_irf branch from master to 'rebase' without divergence


jsnel:
Taking over PR merge responsibilities as discussed @joernweissenborn

PR #270, opened 2019-11-25
Fix generator exit part two

jsnel:
IMPORTANT: DO NOT MERGE - HAS TESTCOMMITS

PR #315, opened 2020-02-24
Add irf_center to result and fix minor runtime issues

### LINKED ISSUE #271: LinAlg runtime errors in test_spectral_penalties.py
In the test
glotaran/builtin/models/kinetic_spectrum/test/test_spectral_penalties.py

the parameter object is used twice, once in an optimization with and once without penalty. The penalty parameters (presumably because unfixed) in the optimization call without penalty causes a LinAlg SingularMatrix error (which is caught so not observed during testing). 

Can be solved by doing a `parameter2 = copy.deepcopy(parameter)` and then `del(parameter2['pen'])`

jsnel:
Add irf_center and irf_width to result object to facilitate plotting

Addresses two generator exist run time warnings. See PR #270 
Fix for LinAlg runtime error. Closes #271

### DIFF
--- a/glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_result.py
+++ b/glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_result.py
@@ -4,6 +4,7 @@ import xarray as xr
 
 import glotaran
 from glotaran.parameter import ParameterGroup
+from glotaran.builtin.models.kinetic_image.irf import IrfMultiGaussian
 from glotaran.builtin.models.kinetic_image.kinetic_image_result import (
     retrieve_species_assocatiated_data,
     retrieve_decay_assocatiated_data,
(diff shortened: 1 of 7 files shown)


glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_result.py
```diff
@@ -53,6 +54,13 @@ def finalize_kinetic_spectrum_result(
         retrieve_decay_assocatiated_data(model, dataset, dataset_descriptor, "spectra")
 
         irf = dataset_descriptor.irf
+        if isinstance(irf, IrfMultiGaussian):
+            if isinstance(irf.center, list):
+                dataset['irf_center'] = irf.center[0].value
```
joernweissenborn: Why not store the full list?

jsnel: In the word of Raymond Hettinger, "There must be a better way". :)

As discussed, a better way is to address the core issue, which is the missing parameters from the result object. And further, the issue of not being able to reproduce the analysis from the output of the saved nc file.

PR #413, opened 2020-09-05
Fixed bug in calling additional penalty function in ungrouped problems

### LINKED ISSUE #399: [bug] equal area constraint implementation
* glotaran version: 0.1.0

### Description of the issue

Currently the equal area constraint (implemented via a additional_penatly_function registered to the model) is calculating an additional penalty per index of the global_axis. See:
https://github.com/glotaran/pyglotaran/blob/master/glotaran/analysis/residual_calculation.py#L30-L35 as part of create_index_independend_ungrouped_residual  
https://github.com/glotaran/pyglotaran/blob/master/glotaran/analysis/residual_calculation.py#L67-L72 as part of create_index_dependend_ungrouped_residual
The  create_index_independend_grouped_residual and create_index_dependend_grouped_residual may have a slightly different implementation. 

Subsequently the amplitude of the clps at an index is plugged into the difference formula, creating a penalty function that penalizes any deviation (amplitude *and* location) rather one that only penalizes a difference in area (amplitude and location may be different). What should happen is that first all the clps are calculated and then (at the same level as the for loop) the penalties are calculated. 

### Additional remarks

As a positive side-effect this bug fix would also create reduces the number of dask.delayed calls.

Currently the spec is implemented as follows:
"equal_area_penalties": [
        {
            "compartment": "s1",
            "target": "s3",
            "parameter": "pen.1",
            "interval": equ_interval,
            "weight": weight,
        },
    ]

Two issues: 
one the interval specified here should be in the global_axis dimension (e.g. wavelength) not the index of the global_axis.
ideally the interval for the (source)compartment and the target(compartment) is specified as two separate ranges, even if these are not of equal length, since the *area* is calculated the penalty can still be caluclated.  

### What I Did

Script to get insight into the problem:
https://gist.github.com/jsnel/a7fdf6b9e972f2c80e1fc6936e9d62f8

joernweissenborn:
Fixes #399


glotaran/analysis/residual_calculation.py
```diff
@@ -37,7 +37,7 @@ def create_index_independent_ungrouped_residual(
             if callable(scheme.model.has_additional_penalty_function):
```
jsnel, line 37: We still need to bring this penalty function calculation at the same level of the preceeding for loop, after all the clps are calculated.

glotaran/analysis/residual_calculation.py
```diff
@@ -82,14 +80,12 @@ def apply_weight(matrix, weight):
...
 
-            if (
-                callable(scheme.model.has_additional_penalty_function)
-                and scheme.model.has_additional_penalty_function()
-            ):
-                additional_penalty = dask.delayed(scheme.model.additional_penalty_function)(
-                    parameter, clp_label, clp, i
-                )
-                penalties.append(additional_penalty)
+            if callable(scheme.model.has_additional_penalty_function):
```
jsnel, line 83: This call needs to not happen inside the for loop, but at  the same level, after the clps are calculated. It then uses the clps and the penalty model spec to calculate the area difference (area1 - penalty.parameter*area2)*penatly.weight and adds that (single number per penalty) to the residual vector.

s-weigand:
@jsnel Would you be so kind to rereview this PR as preparation for #442 ?

glotaran/analysis/residual_calculation.py
```diff
@@ -14,14 +19,14 @@ def create_index_independent_ungrouped_residual(
...
-        size = problem_bag[label].global_axis.size
         weight = problem_bag[label].weight
+        global_axis = problem_bag[label].global_axis
         reduced_clp_labels[label] = constraint_labels_and_matrices[label].clp_label
         matrix = constraint_labels_and_matrices[label].matrix
 
         reduced_clps[label] = []
         residuals[label] = []
-        for i in range(size):
+        for i, index in enumerate(global_axis):
```
s-weigand, line 29: Using `enumerate` and NOT using index doesn't make sense. Also, it is a code smell.

glotaran/analysis/residual_calculation.py
```diff
@@ -60,18 +70,16 @@ def apply_weight(matrix, weight):
...
         data = problem_bag[label].data
-        size = problem_bag[label].global_axis.size
+        global_axis = problem_bag[label].global_axis
         matrices = matrix_jobs[label]
         weight = problem_bag[label].weight
         reduced_clp_labels[label] = []
         reduced_clps[label] = []
         residuals[label] = []
-        for i in range(size):
+        for i, index in enumerate(global_axis):
```
s-weigand, line 79: Using `enumerate` and NOT using index doesn't make sense. Also, it is a code smell.

glotaran/__init__.py
```diff
@@ -4,7 +4,7 @@
 from . import model  # noqa: F401
 from . import parameter
 
-__version__ = "0.1.0"
+__version__ = "0.2.0"
```
jsnel, line 7: Note to self: in the future don't upgrade version numbers within a PR but leave that for the main branch to decide on

jsnel (approved):
There may still be a minor issue remaining in the area of weighted fits, but I suggest patching that outside of this PR.

PR #417, opened 2020-09-07
fix apply_weights call prepare_data from optimize_task

### LINKED ISSUE #416: prepare_data is never called so weights are not applied

jsnel:
cherry picked some changes from the irf_convolution branch

**Testing**

Should pass test

**Closing issues**

closes #416

### DIFF
--- a/glotaran/analysis/optimize.py
+++ b/glotaran/analysis/optimize.py
@@ -40,6 +40,7 @@ def optimize(scheme, verbose=True, client=None):
 
 def optimize_task(initial_parameter, scheme, verbose):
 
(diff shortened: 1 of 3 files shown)


jsnel:
Part of the code from #261 has been pulled into this branch.

s-weigand:
Also not knowing anything about the math and physics behind it YET, after @jsnel  explained to me what the problem was (mostly code not being called) it looks good to me.  I trust @jsnel that the math and physics are fine 😉 

**DISCLAIMER**
Not yet suited for reviews outside of tooling issues.

PR #425, opened 2020-09-30
Add pre-commit autoupdate workflow

s-weigand:
This adds a workflow which runs `pre-commit autoupdate` on schedule or manually triggered with `workflow_dispatch`, and makes a PR if there are new versions of tool.

@jsnel In order for this to trigger the CI at PR's you need [create an access token](https://github.com/technote-space/create-pr-action#github_token) and add it as secret `ACTION_TRIGGER_TOKEN` to the repo.


jsnel (approved):
Secret ACTION_TRIGGER_TOKEN added in the glotaran organization

PR #442, opened 2020-11-20
Refactor/optimization

joernweissenborn:
This PR fill Refactor the Analysis backage and superceds feature/399


s-weigand:
Since this is based on `feature/399`, could you make a PR for that first, and rebase this branch on master after merging?
Especially since @jsnel did some work on `feature/399` this would simplify the reviewing process.

Quick sidenotes:
- `typing.<type_name>` doesn't provide any advantage and using `from typing import <type_name>` leads to a better readability
- There is a new and better way to define namedtuples, which allows typing the slots and providing default values
  Old way:
  ```python 
  from collections import namedtuple
  MyNamedTuple= namedtuple("MyNamedTuple", "slot1 slot2")
  ```
  New way (`python >= 3.6.1`)
  ```python 
  from typing import NamedTuple
  class MyNamedtuple(NamedTuple):
      slot1: int
      slot2: str = "default value"
  ```

- Custom errors
  As per naming convention
  ```python
  ParameterError = ValueError("Parameter not initialized")
  ```
  Should be 
  ```python
  parameter_error = ValueError("Parameter not initialized")
  ```
  Since it's an instance and not a class.
  A proper custom error  would look like this
  ```python
  class ParameterError(ValueError):
      def __init__(self):
           super().__init__("Parameter not initialized")
  ```
  That way the traceback will actually show `ParameterError` and not just `ValueError`.

  Also, this is more flexible since you could use string formatting, e.g. [see](https://github.com/s-weigand/flake8-nb/blob/3db7eb93b224cf663a53f31a3f1338850253708c/flake8_nb/parsers/cell_parsers.py#L35-L56 )

s-weigand (changes requested):
See https://github.com/glotaran/pyglotaran/pull/442#issuecomment-731253232

glotaran/analysis/residual_calculation.py
```diff
@@ -14,14 +19,14 @@ def create_index_independent_ungrouped_residual(
...
-        size = problem_bag[label].global_axis.size
         weight = problem_bag[label].weight
+        global_axis = problem_bag[label].global_axis
         reduced_clp_labels[label] = constraint_labels_and_matrices[label].clp_label
         matrix = constraint_labels_and_matrices[label].matrix
 
         reduced_clps[label] = []
         residuals[label] = []
-        for i in range(size):
+        for i, index in enumerate(global_axis):
```
s-weigand, line 29: Using `enumerate` and NOT using index doesn't make sense. Also, it is a code smell.

glotaran/analysis/residual_calculation.py
```diff
@@ -60,18 +70,16 @@ def apply_weight(matrix, weight):
...
         data = problem_bag[label].data
-        size = problem_bag[label].global_axis.size
+        global_axis = problem_bag[label].global_axis
         matrices = matrix_jobs[label]
         weight = problem_bag[label].weight
         reduced_clp_labels[label] = []
         reduced_clps[label] = []
         residuals[label] = []
-        for i in range(size):
+        for i, index in enumerate(global_axis):
```
s-weigand, line 79: Using `enumerate` and NOT using index doesn't make sense. Also, it is a code smell.

glotaran/analysis/residual_calculation.py
s-weigand: Opps this was intended for #413

glotaran/analysis/residual_calculation.py
s-weigand: Opps this was intended for #413

s-weigand:
I rebased the PR for you, but QA tests are failing since as of #445 we don't allow changes which decrease code documentation.
Check the [new docstring section](https://pyglotaran.readthedocs.io/en/latest/contributing.html#docstrings) for help on how documentation should look like and which tools you can use.

s-weigand:
For type checking you can use
```console
$ mypy  --ignore-missing-imports --scripts-are-modules glotaran/analysis/problem.py
```

s-weigand:
Also, please us a [`postponed annotations`](https://stackoverflow.com/a/61544901/3990615) pattern ([see](https://github.com/glotaran/pyglotaran/pull/453/files#diff-fc1e7b8e8325d1d3f6573de23fbad4f4b7e5d375940355abc0731ed14c5100bd)), to prevent circular dependencies.

All imports that are only used for typing should be inside an `if TYPE_CHECKING:` block.
This will also improve performance since the imports don't need to be done at runtime.

glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_result.py
```diff
@@ -11,59 +8,30 @@
...
     retrieve_species_assocatiated_data,
 )
-from glotaran.parameter import ParameterGroup
 
-from .spectral_constraints import OnlyConstraint
-from .spectral_constraints import ZeroConstraint
 from .spectral_irf import IrfGaussianCoherentArtifact
 from .spectral_irf import IrfSpectralMultiGaussian
 
+T_Problem = typing.TypeVar("glotaran.analysis.problem.Problem")
```
s-weigand, line 15: This will cause 'not valid as a type' see #456 for a reference how it should be done.

s-weigand (changes requested):

glotaran/analysis/problem.py
```diff
@@ -0,0 +1,840 @@
...
+import xarray as xr
+
+from glotaran.analysis.nnls import residual_nnls
+from glotaran.analysis.variable_projection import residual_variable_projection
+from glotaran.parameter import ParameterGroup
+
+from .scheme import Scheme
+
+T_DatasetDescriptor = typing.TypeVar("glotaran.model.DatasetDescriptor")
+T_Model = typing.TypeVar("glotaran.model.Model")
```
s-weigand, line 15: This will cause 'not valid as a type' see #456 for a reference how it should be done.

s-weigand (changes requested):

glotaran/builtin/models/kinetic_spectrum/kinetic_spectrum_result.py
s-weigand: I fixed this one myself after rebasing

s-weigand:
A sidenote on typing with classes:
- `MyClass` refers to an instance of `MyClass` or a subclass of it, same as `isinstance`
- `Type[Myclass]` refers to the `MyClass` or a subclass of it directly and is incompatible with instances of it

jsnel (approved):
All approved - looked at all changes individually as well.
Personally resolved any merge conflicts. 
Good to go.
BusFactor++

s-weigand (approved):
Just trusting @jsnel 's review at this 😄

PR #448, opened 2020-11-25
Pin numpy to 1.19.3 on windows

s-weigand:
This a workaround for an issue that is caused by a windows update (2004).
See https://github.com/numpy/numpy/issues/16744

Since 'numpy==1.19.4' this raises an error  on affected systems when `numpy` is imported.

[Microsoft promised to fix this issue somewhen in January 2021](https://developercommunity.visualstudio.com/comments/1243113/view.html)


jsnel (approved):
Go for it

PR #449, opened 2020-11-25
Raise min required python to 3.8

s-weigand:
To clean up some circular dependency issues we will have to use [`postponed annotations`](https://stackoverflow.com/a/61544901/3990615) which requires python 3.7 .

Since python 3.8 is the current stable and according to @jsnel also the version you get from the windows store (with the major part of this packages audience being windows users), we can take full advantage of the new features in Python 3.8


CONTRIBUTING.rst
```diff
@@ -113,7 +113,7 @@ Before you submit a pull request, check that it meets these guidelines:
 1. The pull request should include tests.
 2. If the pull request adds functionality, the docs should be updated. Put
    your new functionality into a function with a `docstring`_.
-3. The pull request should work for Python 3.6, 3.7. and 3.8
+3. The pull request should work for Python 3.8
```
jsnel, line 116: "should work for Python 3.8, and higher (stable) versions of Python."

jsnel: but, agree, we can add the specific version numbers when we also officially have support for those future Python versions

jsnel (approved):
Ok to pin Python 3.8 as the minimum supported version.

jsnel:
Motivation for PR, other than the use of functionality. 

Python 3.6 is practically End of Life, Python 3.7 only receives security updates, Python 3.8 is still actively receiving bug fixes.
https://devguide.python.org/#status-of-python-branches
Also, Python 3.8 is the stable version in the Windows App Store, as well as the version in the latest Ubuntu LTS.

PR #450, opened 2020-11-25
Add python 3.9 support

s-weigand:
This is PR is based on #449 and as a preparation for when numba adds python 3.9 compatibility, which should be in [about 3 weeks](https://github.com/numba/numba/issues/6345#issuecomment-721738251)


jsnel:
Looks like numba has Python 3.9 support (except perhaps on some newer fruity ARM devices ^^), so lets have another go at enabling Py 3.9 after a rebase @s-weigand .

s-weigand:
@jsnel Numba has merged the patch, but they didn't make a release yet.

jsnel (approved):
⬆️ ⬆️

PR #444, opened 2020-11-21
🔧Use mypy for typechecking

s-weigand:
For the `Decorated property not supported` issue, refer to https://github.com/python/mypy/issues/1362
Locally you can run this check with `pre-commit run -a mypy`, this will currently give you the 233 error the CI shows.

PR #453, opened 2020-11-25
Refactor glotaran.model

s-weigand:
Restructured model package, not to overwrite modules with members ([see](https://github.com/glotaran/pyglotaran/pull/444#issuecomment-733197804)) and fixed a circular dependency issue caused by it.

This reduces the type checking errors from [231](https://github.com/glotaran/pyglotaran/pull/444/checks?check_run_id=1453400975) to [199](https://github.com/s-weigand/pyglotaran/runs/1453403849?check_suite_focus=true).


glotaran/parse/register.py
```diff
@@ -1,7 +1,12 @@
 """A register for models"""
-import typing
+from __future__ import annotations
 
-from glotaran.model.model import Model
+from typing import TYPE_CHECKING
```
jsnel, line 4: Looks a bit odd, but I guess if it's a standard pattern, just go for it.

glotaran/parse/register.py
```diff
@@ -1,7 +1,12 @@
...
+
+if TYPE_CHECKING:
+    from typing import List
+
+    from glotaran.model import Model
```
jsnel, line 9: Looks a bit odd, but I guess if it's a standard pattern, just go for it.

jsnel (approved):
Looks good. Improves readability of the code for me.

glotaran/parse/register.py
s-weigand: Since the import is only used for type checking, it kinda makes sense to only import it when actually type checking.

joernweissenborn:
I approve under the condition that @jsnel  will do the rebase of the refactor/optimization branch.

joernweissenborn (approved):

PR #456, opened 2020-11-26
Fix 'not valid as a type' type checking errors

s-weigand:
This PR fixes all 'not valid as a type' type errors mypy trows and makes some module member exports more pythonic.

It reduces the type checking errors from [199](https://github.com/glotaran/pyglotaran/pull/444/checks?check_run_id=1458384107) to [183](https://github.com/s-weigand/pyglotaran/runs/1459302847?check_suite_focus=true).

In some cases `any` (builtin function) was used as type of index so I replaced them with of `typing.Any`.
But I think it should be `int` or `Union[int, float]`.
Also since the implementation of the interval checking is always the same it can be refactored to a helper function.

And
```python
def applies(interval):
    return interval[0] <= index <= interval[1]
```
should be 
```python
def applies(interval):
    return min(interval) <= index <= max(interval)
```
or there should be some validation e.g.
```python
if not math.isclose(interval[0], min(interval)):
    raise ValueError(f"Interval needs to be ordered. Got {interval}")
```

### DIFF
--- a/glotaran/builtin/models/kinetic_spectrum/spectral_constraints.py
+++ b/glotaran/builtin/models/kinetic_spectrum/spectral_constraints.py
@@ -1,21 +1,25 @@
 """This package contains compartment constraint items."""
+from __future__ import annotations
 
-import typing
-
-import numpy as np
+from typing import TYPE_CHECKING
+from typing import List
+from typing import Tuple
 
 from glotaran.model import model_attribute
 from glotaran.model import model_attribute_typed
 
-T_KineticSpectrumModel = typing.TypeVar(
-    "glotaran.builtin.models.kinetic_spectrum.KineticSpectrumModel"
-)
+if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


glotaran/builtin/models/kinetic_spectrum/spectral_constraints.py
```diff
@@ -49,7 +53,7 @@ def applies(interval):
 @model_attribute(
     properties={
         "compartment": str,
-        "interval": typing.List[typing.Tuple[float, float]],
+        "interval": List[Tuple[float, float]],
```
s-weigand, line 56: If that type is correct, than there is no point in checking
```python
if isinstance(self.interval, tuple):
```
in the implementation.
Else the type should be `Union[Tuple[float, float], List[Tuple[float, float]]]`.

jsnel (approved):
All seems ok to me.

PR #464, opened 2020-11-29
Fix strange behaviour of interrogate hook

s-weigand:
Now interrogate should always run on the whole project, as intended.

This is a [known issue](https://github.com/econchick/interrogate/issues/60) when running as pre-commit hook.

Since my projects where I use it have 100% coverage I didn't see that before, so sorry for the problem 😢


.pre-commit-config.yaml
```diff
@@ -43,7 +43,7 @@ repos:
     hooks:
       - id: interrogate
         args: [-vv, --config=pyproject.toml]
-        exclude: ^(docs|setup.py)|test/|__init__.py
+        pass_filenames: false
```
jsnel, line 46: documentation here: https://pre-commit.com/#hooks-pass_filenames

jsnel (approved):
LGTM

.pre-commit-config.yaml
s-weigand: This change makes `interrogate` basically behave like if you would call it from the command line.

And the whole idea behind using `interrogate` was, put simply: "Don't add new code if it isn't documented!"
Basically a test coverage for docstrings.

So at some point (close to v1 release) we can use [pydocstyle](https://github.com/PyCQA/pydocstyle) and [darglint](https://github.com/terrencepreilly/darglint) to have a good and in sync API documentation.

PR #465, opened 2020-11-29
Remove legacy python versions as target from black config

s-weigand:
Some artifacts from before #449


jsnel (approved):
As discussed. Thanks.

PR #470, opened 2020-12-01
Remove leftover dask artifacts

s-weigand:
Removed leftover dask artifacts in dependencies since dask isn't part of glotaran since merging #455


jsnel (approved):
OK

Issue #478, opened 2020-12-06
In two dataset (grouped) analysis self._full_clp_labels is None


jsnel:
* glotaran version: v0.2.0 (312e075026f26ef4490c747a958d27ff6cb7e4b7 from PR477)
* Python version: 3.8.5
* Operating System: Windows 10

### Description

First patch with PR477 since otherwise another crash is observed prior to reaching the point of the traceback below.
When running a [(two) dataset analysis](https://github.com/glotaran/pyglotaran_examples/blob/master/examples/two_dataset/two_dataset_analysis.py) with [equal_area_penalties in the model](https://github.com/glotaran/pyglotaran_examples/blob/master/examples/two_dataset/models/equareasim2_model.yaml) a crash is observed due clp_labels being None
Two dataset so grouped, call to self.calculate_index_dependent_grouped_residual()
In the call of _calculate_additional_grouped_penalty  self._full_clp_labels is None
### What I Did

```
Traceback (most recent call last):
  File "c:/src/pyglotaran_examples/examples/two_dataset/two_dataset_analysis.py", line 60, in <module>
    result = optimize(scheme)
  File "c:\src\pyglotaran\glotaran\analysis\optimize.py", line 27, in optimize
    lm_result = minimizer.minimize(method="least_squares", verbose=verbose, max_nfev=scheme.nfev)
  File ".envs\py38gta_pdev\lib\site-packages\lmfit\minimizer.py", line 2285, in minimize
    return function(**kwargs)
  File ".envs\py38gta_pdev\lib\site-packages\lmfit\minimizer.py", line 1527, in least_squares
    ret = least_squares(self.__residual, start_vals,
  File ".envs\py38gta_pdev\lib\site-packages\scipy\optimize\_lsq\least_squares.py", line 807, in least_squares
    f0 = fun_wrapped(x0)
  File ".envs\py38gta_pdev\lib\site-packages\scipy\optimize\_lsq\least_squares.py", line 802, in fun_wrapped
    return np.atleast_1d(fun(x, *args, **kwargs))
  File ".envs\py38gta_pdev\lib\site-packages\lmfit\minimizer.py", line 576, in __residual
    out = self.userfcn(params, *self.userargs, **self.userkws)
  File "c:\src\pyglotaran\glotaran\analysis\optimize.py", line 48, in _calculate_penalty
    return problem.full_penalty
  File "c:\src\pyglotaran\glotaran\analysis\problem.py", line 239, in full_penalty
[...]

PR #479, opened 2020-12-06
Small optimize and problem class refactor

jsnel:
Added optimizer call for the problem class
Refactored problem class to change name of full_clps to just clps

This is only partial work towards fixing issue #478, but this needs to be in prior to further refactoring.


s-weigand (approved):
LGTM

PR #493, opened 2021-01-10
Replace lmfit with scipy.optimize

### LINKED ISSUE #481: Parsing parameter relations with  numexpr or sympy
### Description

@joernweissenborn told me that one of the main points to keep `lmfit` (with all the known issues we have with it), is the parsing of parameter relations. Depending on the specs we need there are multiple solutions (all of which will need some string preprocessing, due to the dot syntax e.g. `rates.k1` would need to be changed with a simple string replace to `rates_k1` since it needs to be a valid python variable name, this is ofc just under the hood).

If we just need to evaluate expressions we could use [numexpr](https://pypi.org/project/numexpr/).

If we need to transform a string equation to a function we could use a combination of sympy functions i.e. [`sympify`](https://docs.sympy.org/latest/modules/core.html#sympy.core.sympify.sympify) and [`lambdify`](https://docs.sympy.org/latest/modules/utilities/lambdify.html#sympy.utilities.lambdify.lambdify)

Here is a basic example of how to generate an arbitrary numerical function from a string with sympy
```python
In [1]: import sympy as sym

In [2]: expr = sym.sympify("a*exp(2*x)")

In [3]: expr.free_symbols
Out[3]: {a, x}

In [4]: expr_func = sym.lambdify(expr.free_symbols, expr=expr, modules="numpy")

In [5]: expr_func(x=0, a=5)
Out[5]: 5.0

In [6]: expr_func(a=5, x=0)
Out[6]: 5.0

In [7]: import numpy as np

In [8]: expr_func(a=5, x=np.arange(0, 10))
Out[8]:
array([5.00000000e+00, 3.69452805e+01, 2.72990750e+02, 2.01714397e+03,
       1.49047899e+04, 1.10132329e+05, 8.13773957e+05, 6.01302142e+06,
       4.44305526e+07, 3.28299846e+08])
```

An added bonus with `sympy` is expression simplification and latex output e.g. for automated report generation (maybe that would be a nice to have for `pyglotaran_extras` 😉 ).

### LINKED ISSUE #489: Switch optimiztion to use scipy.optimize directly instead of lmfit

### LINKED ISSUE #491: Create regular expression to validate parameter names and and capture group strings with '$' in user defined expressions for parameter values.
# Needed regular expressions

1. Validate parameter names
2. Capture references in user defined relation expressions (for inline replacement)

## Regex to validate parameter names 
 e.g. 2. "my_parameter1" or simply "1" 

##  Regex to capture references in user-defined expressions
 e.g.: I have an expression like "$kinetic.1 * exp($kinetic.2)" for example, I want to transform into an expression like "group.get(kinetic.1') * exp(group.get('kinetic.2')"

joernweissenborn:
This PR removes all dependencies to the `lmfit` package.

The optimization is now done by directly calling `scipy.optimize.least_squares`.

Changes
=======

* The `Parameter` class is no longer a child of `lmfit.Parameter`.
* `asteval` is used to evaluate parameter expressions.
* `optimize` now calls `scipy.optimize.least_quares` instead of using `lmfit.Minimizer`
* Few adaptions to tests due to small numeric differences
* Parameter and group labels now have restrictions to work with the new parameter expression scheme

Known Issues
===========

* The covariance matrix is no longer calculated
* The standart error of the optimized parameters is not calculated

This PR closes #491, closes #489 and closes #481.

### DIFF
--- a/glotaran/builtin/models/kinetic_spectrum/test/test_spectral_irf.py
+++ b/glotaran/builtin/models/kinetic_spectrum/test/test_spectral_irf.py
@@ -280,7 +280,7 @@ def test_spectral_irf(suite):
     assert np.array_equal(dataset["spectral"], resultdata["spectral"])
     assert dataset.data.shape == resultdata.data.shape
     assert dataset.data.shape == resultdata.fitted_data.shape
-    assert np.allclose(dataset.data, resultdata.fitted_data, atol=1e-15)
+    assert np.allclose(dataset.data, resultdata.fitted_data, atol=1e-14)
 
     print(resultdata.fitted_data.isel(spectral=0).argmax())
     print(resultdata.fitted_data.isel(spectral=-1).argmax())
--- a/glotaran/builtin/models/kinetic_spectrum/test/test_spectral_relations.py
+++ b/glotaran/builtin/models/kinetic_spectrum/test/test_spectral_relations.py
@@ -113,6 +113,10 @@ def test_spectral_relation():
     scheme = Scheme(model=model, parameter=parameter, data=dataset, nfev=20)
     result = optimize(scheme)
 
+    for label, param in result.optimized_parameter.all():
+        if param.vary:
+            assert np.allclose(param.value, parameter.get(label).value, rtol=1e-1)
+
     result_data = result.data["dataset1"]
     print(result_data.species_associated_spectra)
     assert result_data.species_associated_spectra.shape == (1, 4)
@@ -120,9 +124,9 @@ def test_spectral_relation():
         result_data.species_associated_spectra[0, 1]
         == rel1 * result_data.species_associated_spectra[0, 0]
     )
-    assert (
-        result_data.species_associated_spectra[0, 2]
-        == rel2 * result_data.species_associated_spectra[0, 0]
+    assert np.allclose(
+        result_data.species_associated_spectra[0, 2].values,
+        rel2 * result_data.species_associated_spectra[0, 0].values,
     )
 
 
--- a/glotaran/parameter/parameter_group.py
+++ b/glotaran/parameter/parameter_group.py
@@ -378,7 +378,7 @@ class ParameterGroup(dict):
     def set_from_label_and_value_arrays(self, labels: List[str], values: np.ndarray):
         """Updates the parameter values from a list of labels and values."""
 
-        if not len(labels) == len(values):
+        if len(labels) != len(values):
             raise ValueError(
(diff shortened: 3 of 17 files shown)


glotaran/builtin/models/kinetic_spectrum/test/test_spectral_irf.py
```diff
@@ -280,7 +280,7 @@ def test_spectral_irf(suite):
     assert np.array_equal(dataset["spectral"], resultdata["spectral"])
     assert dataset.data.shape == resultdata.data.shape
     assert dataset.data.shape == resultdata.fitted_data.shape
-    assert np.allclose(dataset.data, resultdata.fitted_data, atol=1e-15)
+    assert np.allclose(dataset.data, resultdata.fitted_data, atol=1e-14)
```
jsnel, line 283: Yes, this tiny raising of absolute tolerance is acceptable. 

But this test needs to be revised anyways, when I do that I will have another look.

glotaran/builtin/models/kinetic_spectrum/test/test_spectral_relations.py
```diff
@@ -113,16 +113,20 @@ def test_spectral_relation():
     scheme = Scheme(model=model, parameter=parameter, data=dataset, nfev=20)
     result = optimize(scheme)
 
+    for label, param in result.optimized_parameter.all():
+        if param.vary:
+            assert np.allclose(param.value, parameter.get(label).value, rtol=1e-1)
```
jsnel, line 118: Will make a note to revisit proper relative tolerance to be used here, see #495.

glotaran/parameter/parameter_group.py
```diff
@@ -202,10 +164,10 @@ def from_csv(cls, filepath: str, delimiter: str = "\t"):
                 p = Parameter(label=label.pop())
                 p.value = df["value"][i]
                 p.stderr = df["stderr"][i]
-                p.min = df["min"][i]
-                p.max = df["max"][i]
+                p.minimum = df["min"][i]
```
jsnel, line 167: Note: we're not changing the key for csv with this PR, hence renaming the key for internal use only.

glotaran/parameter/test/test_parameter.py
```diff
@@ -1,5 +1,9 @@
+import pickle
```
jsnel, line 1: We have considered the security implication of using pickle here (for testing only, in a safe way).

jsnel (approved):
LGTM!

PR #500, opened 2021-01-16
Add CODEOWNERS file to main branch

jsnel:
Added CODEOWNERS for clarifying primary code responsibility

Following documentation here:
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -1,0 +1,20 @@
+name: "Codeowners Validator"
+
+on:
+  schedule:
+    # Runs at 08:00 UTC every day
+    - cron:  '0 8 * * *'
+
+jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      # Checks-out your repository, which is validated in the next step
+      - uses: actions/checkout@v2
+      - name: GitHub CODEOWNERS Validator
+        uses: mszostok/codeowners-validator@v0.5.1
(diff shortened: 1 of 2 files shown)


jsnel:
@s-weigand in addition to the CODEOWNERS file, this [codeowners-validator github action](https://github.com/mszostok/codeowners-validator) would perhaps be interesting to add to the list of checks?

.github/workflows/sanity.yml
```diff
@@ -0,0 +1,20 @@
+name: "Codeowners Validator"
+
+on:
+  schedule:
+    # Runs at 08:00 UTC every day
+    - cron:  '0 8 * * *'
```
s-weigand, line 6:
```suggestion
    - cron:  '0 8 * * *'
  workflow_dispatch:
```
I came to love the `workflow_dispatch` trigger, especially for scheduled workflows, since it allows you to trigger the workflow whenever you want.
That said it might also be good to run it on `push` and/or `pull_request` in case the `CODEOWNERS` file gets changed e.g. if we add a new subpackage with specific owners.
Since the workflow isn't coupled to other steps and should be pretty quick this shouldn't produce a big overhead.

s-weigand (dismissed):
A nice addition to the tooling chain ❤️

joernweissenborn (approved):

PR #502, opened 2021-01-17
Support parameter expression across parameter groups

### LINKED ISSUE #501: Fix parameter expression parsing when using groups
Parameter expression parsing currently only works if all parameter live in the root node, not when they are organized in groups.

When a parameter in group A is referencing a paramter in group B an error is thrown

> ValueError: Expression '1 - $B.1' of parameter '2' evaluates tonon numeric value 'None'.

Also there is a typo in the error, it should be: "... to non-numeric ..."

jsnel:
Primary goal of this PR:

- [ ] Support for parameter expression parsing across different groups.

Tasks:

- [x] Added unit test to demonstrate test failure
- [ ] Fix expression parsing when using groups, update unit tests 
- [ ] Pass unit tests

**Closing issues**

closes #501

### DIFF
--- a/glotaran/parameter/test/test_parameter.py
+++ b/glotaran/parameter/test/test_parameter.py
@@ -284,10 +284,51 @@ def test_parameter_expressions():
 
     with pytest.raises(ValueError):
         params_bad_expr = """
-    - ["3", {expr: 'None'}]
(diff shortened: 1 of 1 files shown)


jsnel:
Close in favor of #504

PR #508, opened 2021-01-21
Fix CODEOWNERS file

s-weigand:
There were some issues with the `CODEOWNERS` file which rendered it invalid.
This PR fixes the issues (well at least the [test workflow now passes](https://github.com/glotaran/pyglotaran/runs/1742944684?check_suite_focus=true) 😉 )
I had to open a branch here instead of using my fork, since:
a) I needed access to `secrets.OWNERS_VALIDATOR_GITHUB_SECRET`
b) The API request path resolves based on which repo it is run on

Issues were:
- not fully qualified orga-group-name 
- typo in a path
- overpopulated file patterns (only one per line allowed)

Other changes:
- renamed the workflow so it is more obvious what it does
- made workflow run on push and only in the glotaran organization


.github/CODEOWNERS
```diff
@@ -15,18 +15,23 @@
...
+/*.y*ml                         @glotaran/admins @glotaran/maintainers
+/*.ini                          @glotaran/admins @glotaran/maintainers
+/*.toml                         @glotaran/admins @glotaran/maintainers
+/*.txt                          @glotaran/admins @glotaran/maintainers
+LICENSE                         @glotaran/pyglotaran_creators
 
 # docs
-/docs *.rst *.md                @glotaran/maintainers @pyglotaran_creators
+/docs/**/*.rst                  @glotaran/maintainers @glotaran/pyglotaran_creators
+# /docs/**/*.md                   @glotaran/maintainers @glotaran/pyglotaran_creators
```
jsnel, line 28: Why is this left in a a comment?

jsnel (approved):
LGTM!

.github/CODEOWNERS
s-weigand: Just for when we add `*.md` files to the docs.
With the [myst-parser](https://myst-parser.readthedocs.io/en/latest/using/syntax.html) you get the power of `*.rst` in `*.md` 😉

PR #522, opened 2021-01-29
Refactored Scheme

### LINKED ISSUE #521: Improvement to prepare_data for data weight application
* glotaran version: [4411e6de8d](https://github.com/glotaran/pyglotaran/blob/4411e6de8d07d2fc806514bce139619d4c3b2c44)

### Description

Currently data weight are not applied unless prepare_data is explicitly called. 
 
Ensure [prepare_data](https://github.com/glotaran/pyglotaran/blob/4411e6de8d07d2fc806514bce139619d4c3b2c44/glotaran/analysis/scheme.py#L197) is always called since it is crucial for data weights application.

And there is room for further improvement to this orphaned function.

Tests by default should not need to call this function directly.

joernweissenborn:
Fixes #521 

IMPORTANT
=======

Changed argument and property names of  `Scheme` to comply with glotaran naming conventions.


jsnel (approved):
When going the more verbose route I would have gone with maximum_number_of_function_evaluations instead of maximum_number_function_evaluations but that can be changed in the next PR.

Issue #532, opened 2021-02-05
Refactor tests to address code duplication


jsnel:
* glotaran version: current master 
* Python version:
* Operating System:

### Description

The [Sonarcloud ](https://sonarcloud.io/dashboard?id=glotaran_pyglotaran&branch=master) bot is failing one the basis of [code duplication exceeding the 3%](https://sonarcloud.io/component_measures?id=glotaran_pyglotaran&metric=new_duplicated_lines_density&view=list) threshold (we're now at 4%), which is mostly due to copy and pasting large dictionaries defining fitting / simulation models. 

### Description
There is a number of solution to this problem.
- Work with nested dicts (derive from a base model dicts and add/del keys as needed in each tests)
- Move specification to .yml files so the bots (largely) ignore them / they don't code towards code duplication.
- Make small (yaml) code generator function to build up a model spec. 
- Make a model spec reusable between fitting and simulation. Allowing a model spec to contain e.g. shapes specification which can be ignore during fitting and can be using during simulation.

jsnel:
Fixed in #533

PR #533, opened 2021-02-05
Refactor tests to address code duplication

jsnel:
See test refactor issue #532

The first commit in this PR illustrates two possible solution to refactoring the tests:
- Move specification to .yml files so the bots (largely) ignore them / they don't code towards code duplication.
- Make small (yaml) code generator function to build up a model spec.
The other two options are left as an exercise to the reader (to be committed in a follow up commit):
- Work with nested dicts (derive from a base model dicts and update/del keys as needed in each tests)
- Make a model spec reusable between fitting and simulation. Allowing a model spec to contain e.g. shapes specification which can be ignore during fitting and can be using during simulation.

The latter solution is preferred by the commit's author.


glotaran/builtin/models/kinetic_spectrum/test/test_spectral_irf.py
```diff
@@ -244,38 +114,27 @@ def test_spectral_irf(suite):
...
-    assert sim_model.valid(wanted_parameters)
-
-    initial_parameters = suite.initial_parameters
-    print(model.validate(initial_parameters))
-    assert model.valid(initial_parameters)
-
-    print(model.markdown(wanted_parameters))
+    parameters = suite.parameters
+    print(model.validate(parameters))
+    assert model.valid(parameters)
```
s-weigand, line 119: I see that you removed the `wanted_parameters` since they were the same as `initial_parameters`, but shouldn't they be different to see if `optimize` works properly?

glotaran/builtin/models/kinetic_spectrum/test/test_kinetic_spectrum_model.py
```diff
@@ -1,382 +1,233 @@
...
-                "dataset1": {
-                    "initial_concentration": "j1",
-                    "irf": "irf1",
-                    "megacomplex": ["mc1"],
-                    "shape": {"s1": "sh1", "s2": "sh2", "s3": "sh3"},
-                },
-            },
-        }
-    )
+MODEL_3C_BASE = """\
```
s-weigand, line 94: For the naming pattern to be consistent with the rest of the code the 3C should be spelled out, e.g.  `MODEL_3C_BASE` should be `MODEL_3_COMPONENT_BASE`.

glotaran/builtin/models/kinetic_spectrum/test/test_kinetic_spectrum_model.py
```diff
@@ -1,382 +1,233 @@
...
 import pytest
 
+from glotaran import read_model_from_yaml
+from glotaran import read_parameters_from_yaml
 from glotaran.analysis.optimize import optimize
 from glotaran.analysis.scheme import Scheme
-from glotaran.builtin.models.kinetic_spectrum import KineticSpectrumModel
-from glotaran.parameter import ParameterGroup
 
+MODEL_1C_BASE = """\
```
s-weigand, line 9: For the naming pattern to be consistent with the rest of the code the `1C` should be spelled out, e.g. `MODEL_1C_BASE` should be `MODEL_1_COMPONENT_BASE`.

s-weigand (changes requested):
Nice changes, less copypasta, and the differences are much more obvious. 😄

glotaran/builtin/models/kinetic_spectrum/test/test_kinetic_spectrum_model.py
jsnel: Although in general I would agree with increased verbosity, in particular in any function or class names, 
this is only a constant used in supporting classes of unit tests. 

After careful consideration (trying different things for a few hours while refactoring) I found this to be the right balance between verbosity, readability and maintainability. Maybe in the future we find a better way. I'll be sure to smash my fist on the table when I do.

glotaran/builtin/models/kinetic_spectrum/test/test_kinetic_spectrum_model.py
jsnel: Although in general I would agree with increased verbosity, in particular in any function or class names, 
this is only a constant used in supporting classes of unit tests. 

After careful consideration (trying different things for a few hours while refactoring) I found this to be the right balance between verbosity, readability and maintainability. Maybe in the future we find a better way. I'll be sure to smash my fist on the table when I do.

glotaran/builtin/models/kinetic_spectrum/test/test_spectral_irf.py
jsnel: Yes, but this is not an optimizer test, and they were also not different in the initial test (the same parameters were used in both cases). I specifically refactored to make that point clear.

jsnel (commented):
LGTM

PR #540, opened 2021-02-07
Fixed arguments in Result's get_scheme call

### LINKED ISSUE #539: Wrong call in get_scheme in result.py
* glotaran version: current master

### Description
LGTM was so helpful to point out this _Wrong name for an argument in a class instantiation_ [in result.py](https://github.com/glotaran/pyglotaran/blob/c72e7b859067dcc10669576a0efc3416baa93f46/glotaran/analysis/result.py#L222-L241)
https://lgtm.com/projects/g/glotaran/pyglotaran/?mode=list
get_scheme
The error was not caught since it's not under unit test.

jsnel:
Fixed code to the point where get_scheme works to return a new scheme object on which optimize can be called.

Closes #539

### DIFF
--- a/glotaran/analysis/result.py
+++ b/glotaran/analysis/result.py
@@ -238,7 +238,7 @@ class Result:
         return Scheme(
             model=self.model,
             parameters=self.optimized_parameters,
-            data=self.data,
+            data=data,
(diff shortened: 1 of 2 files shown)


s-weigand (approved):
LGTM

PR #549, opened 2021-02-11
Refactor/explicit file loader

### LINKED ISSUE #531: Address deprecation warnings
### During run time

One coming from numpy whenever `read_data_file` is called, e.g.:
```
dataset = read_data_file(data_path)
```
> numpy\core\_asarray.py:83: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
>   return array(a, dtype, copy=False, order=order)

### During testing
In test_coherent_artifact.py::test_coherent_artifact
(similar to the one above)

>   numpy\core\_asarray.py:83: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
>     return array(a, dtype, copy=False, order=order)

Also in test_coherent_artifact.py::test_coherent_artifact

> glotaran/builtin/models/kinetic_spectrum/test/test_coherent_artifact.py::test_coherent_artifact
>   c:\users\jsnel\anaconda3\envs\pip38gta\lib\site-packages\numpy\core\numeric.py:2378: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
>     return bool(asarray(a1 == a2).all())

glotaran/builtin/models/kinetic_spectrum/test/test_spectral_irf.py::test_spectral_irf

>    xarray\core\dataarray.py:4244: DeprecationWarning: Behaviour of argmin/argmax with neither dim nor axis argument will change to return a dict of indices of each dimension. To get a single, flat index, please use np.argmin(da.data) or np.argmax(da.data) instead of da.argmin() or da.argmax().
>     result = self.variable.argmax(dim, axis, keep_attrs, skipna)

Observed when running
```
pytest
```

jsnel:
Addresses the last VisibleDeprecationWarning related to casting a ragged list of lists (of unequal length) into a numpy array.
The first list being 1 element less than the following. Problem solved by using pandas for reading in the data, and splitting the reading into two parts.

**Closing issues**

Closes #531


s-weigand (approved):
LGTM

PR #560, opened 2021-02-13
Change setup

### LINKED ISSUE #559: Move setuplogic from setup.py to setup.cfg
### Description
Since our deployment strategy changed to merge a release PR and using github releases to create tags there is no more need to use `bump2version` if we change from `setup.py` to `setup.cfg`

Right now `bump2version` just changes the version string in `setup.py` and `glotaran/__init__.py` and creates a commit + tag.
Setuptools>=39.2.0 introduced the [`attr:` property for `version`](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html?highlight=metadata#metadata) which would let us define which attribute of the package is the version (`glotaran.__version__` in our case), thus we wouldn't need `bump2version` anymore.

Furthermore, using `setup.cfg` would allow us to use a formatted and linter on the setup logic

s-weigand:
This PR moves the setuplogic from `setup.py` to `setup.cfg` and adds the `setup-cfg-fmt` formatting and lining pre-commit-hook.

**Testing**

**Closing issues**

closes #559


jsnel (approved):
OK

PR #556, opened 2021-02-12
Rework IO and result saving

joernweissenborn:

PR #562, opened 2021-02-14
Fix/feature extended result saving

jsnel:
Testing out staticmethod fix with LGTM bot


jsnel:
Experimental PR, work continues on #556

PR #563, opened 2021-02-14
Remove 'docs' artifact from CI, in favor of RTD build on PR

s-weigand:
I set [RTD to build on PR's](https://docs.readthedocs.io/en/latest/pull-requests.html) so we don't need the docs artifact anymore.
This PR is also a test for RTD to build the docs.

**Testing**


jsnel (approved):
Very nice. Reviewed ok.

Inspected the generated docs, looks just like the read thing, with the additional warning that
"The page was created from a pull request."
The build pages can be viewed here:
https://readthedocs.org/projects/pyglotaran/builds/

PR #567, opened 2021-02-21
Update to codeowners

jsnel:
Update to CODEOWNERS file following changes introduced by the merge of PR #556

PR 556 introduced a new IO system introducing new builtin modules and deprecated the parse module as well as builtin/file_formats


s-weigand (approved):
LGTM

PR #577, opened 2021-02-28
Updated warning message in quickstart

jsnel:
Replaced the warning indicating to not use it for science.

Added reference to v0.3 maintenance branch version of the README


s-weigand (approved):
LGTM

s-weigand:
Maybe it would be better to link to the readme page of that version instead of the version on `main` 🤔

PR #590, opened 2021-03-08
Use tabulate for table markdown formatting

### LINKED ISSUE #587: Use tabulate for table markdown formatting (#556 follow up)
Instead of reinventing the wheel (pun semi-intended) and doing the markdown formatting of table-like output ourselves, we should use [tabulate](https://pypi.org/project/tabulate/) which is battle proven (e.g. also used by `pandas`), will reduce "busy work" and improve code readability.

s-weigand:
Changes:
- Adds tabulate as a dependency
- `Result.markdown` now renders properly
- `ParameterGroup.markdown` is easier to read (both code and output)

A demo of how it looks [can be found at binder](https://mybinder.org/v2/gh/s-weigand/pyglotaran/tabulate?urlpath=lab%2Ftree%2Fdocs%2Fsource%2Fnotebooks%2Fquickstart.ipynb).

Compared it to [the current version](https://mybinder.org/v2/gh/glotaran/pyglotaran/main?urlpath=lab%2Ftree%2Fdocs%2Fsource%2Fnotebooks%2Fquickstart.ipynb), where the output of cell 16 showcases all changes.
To see the difference in plaintext (e.g. what a CLI user would see) use `print` instead of `print_md`.

**Testing**

**Closing issues**

closes #587

### DIFF
--- a/glotaran/parameter/parameter_group.py
+++ b/glotaran/parameter/parameter_group.py
@@ -379,9 +379,12 @@ class ParameterGroup(dict):
                 parameter.value = value
 
     def markdown(self) -> str:
-        """Formats the :class:`ParameterGroup` as markdown string."""
-        t = "  " * self.get_nr_roots()
-        s = ""
+        """Formats the :class:`ParameterGroup` as markdown string.
+
+        This is done by recursing the nested :class:`ParameterGroup` tree.
+        """
+        node_indentation = "  " * self.get_nr_roots()
+        return_string = ""
         parameter_rows = []
         table_header = [
             "_Label_",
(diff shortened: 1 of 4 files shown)


glotaran/parameter/parameter_group.py
```diff
@@ -377,16 +379,48 @@ def update_parameter_expression(self):
...
+            "_Label_",
+            "_Value_",
+            "_StdErr_",
+            "_Min_",
+            "_Max_",
+            "_Vary_",
+            "_Non-Negative_",
+            "_Expr_",
+        ]
         if self.label != "p":
```
joernweissenborn, line 398: This line is a leftover from the limitations of lmfit's Parameters. Back then, the root group had the user invisible label `p`, no it is `None`. So the check should be `if label is not None`.

joernweissenborn (changes requested):

joernweissenborn (approved):

jsnel (approved):
Much improved!

PR #595, opened 2021-03-13
Add configs for tools planned to use

s-weigand:
Just some tool configs for QA tools we plan to use but can't for the whole project because then the CI would fail.
When I write code for PR's I check that my code passes tests with `mypy`, `pydocstyle` and `darglint`, which we also plan to use.
So it would be nice to already have the configs in place, so I don't need to pass all the CLI flags each time


jsnel (approved):
👍

jsnel (approved):
👍

PR #596, opened 2021-03-13
Move doc dependencies to a separate file

s-weigand:
That way we don't install unneeded packages (e.g. `pytest`) when testing the docs.
Also, the `doc` jobs are a kind of sanity check if we miss dependencies in our setup.


readthedocs.yml
```diff
@@ -8,6 +8,6 @@ build:
 python:
   version: 3.8
   install:
-    - requirements: requirements_dev.txt
+    - requirements: docs/doc-dependencies.txt
```
jsnel, line 11: As per RTD https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#using-a-configuration-file I suggest changing to docs/requirements.txt

s-weigand: Done :smile:

jsnel (approved):
👍

PR #600, opened 2021-03-16
Refactor plugin system

### LINKED ISSUE #586: Refactor plugin system (#556 follow up)
IMHO the plugin system would profit quite a lot from some conceptual changes, with most of them only visible to core and plugin developers.

### Move all the plugin registration code in one module/package
IMHO it would make more sense to have a `plugin_system` module/package with all the read/write registrations code than having `register.py` modules in `glotaran.io` and `glotaran.model`:
- One place one responsibility
- Less code duplication

### Add warning when a plugin replaces an existing one
Right now (since we only have builtin plugins) we just add a key-value pair to the registry dict.
Since we will have no control over the naming in 3rd party plugins this could possibly lead to overwriting builtin plugin or 3rd party plugins overwriting each other. This problem concerns `model` as well as `Io` plugins

https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/model/register.py#L22

https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/io/register.py#L21

Thus there should be at least a warning if a plugin with the same name is already registered, telling the user which plugin overwrites which other plugin. So at least issues can be directed to the proper code owners.

A possibly more robust option would be to `pre`-/`post`-fix plugin entries and inform the users about the naming change, so they can use the version they want.
E.g. there are two plugins with the name `sdt_file`, one is our version from `builtin` the other one from a fictional package `gta_plugin_sdt2021`.
The first one to be registered (I'm not sure about the ordering of `entry_points`) would be under the name `sdt_file` and the 2nd one `sdt_file_pyglotaran`/`sdt_file_gta_plugin_sdt2021` and the user gets a warning that there was a conflict with plugins originating from the corresponding modules.

See the implementation of plugin registering:
https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/__init__.py#L5-L6

### Splitup entry points to be more granular
Right now the entry points for plugins are one big blob, not differentiating between the plugins responsibilities:
https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/setup.cfg#L52-L59

A more granular structure would improve readability, add the possibility to list installed plugins by their responsibility and reduce possible naming conflicts.
E.g.
```ini
glotaran.plugins.model =
    kinetic_image = glotaran.builtin.models.kinetic_image
    kinetic_spectrum = glotaran.builtin.models.kinetic_spectrum
glotaran.plugins.data_io=
    ascii = glotaran.builtin.io.ascii.wavelength_time_explicit_file
    sdt = glotaran.builtin.io.sdt.sdt_file_reader
    csv = glotaran.builtin.io.csv.csv
    nc = glotaran.builtin.io.netCDF.netCDF
glotaran.plugins.project_io=
    yml = glotaran.builtin.io.yml.yml
```

### Rename `Io` and possibly split it up
The name `Io` doesn't fit well in the verbose naming scheme of `pyglotaran` (I know naming things is damn hard 😉 ), thus a name like `IoInterface` or `IoPlugin` would be more descriptive.

Personally, I would strongly recommend splitting up `Io` in `DataIoInterface` and `ProjectIoInterface` since those plugins have a disjunct responsibility and use case:
- Many people will want to have a data reader for their setups format (e.g. `sdt_file` has no model functionality)
- Only a few will want to change the way how they define their model and parameters (e.g. `yml_io` has no data functionality)
- It will be easier for plugin developers  to spot what they want and need ("single" responsibility)
- If users follow the docs, this will be hidden from them anyway since we got convenience functionalities like `load_model` which just access the corresponding plugin registry
https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/io/register.py#L40

### Make use of the `Io` instances and remove the `fmt` arg
Since we instantiate `Io` plugins when resisting them 
https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/io/register.py#L21
there is no need to have the `fmt` argument in the methods e.g.:
https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/builtin/io/yml/yml.py#L55
since you could just use `self.fmt`.
The functions used by the users e.g. `load_model` just access the registry and pass `fmt` again since it wasn't part of the constructor of the instance. The only way to get the wrong instance for a format would be to directly instantiate the class with the wrong format.

### Add a `get_dataloader` (maybe even `get_datawriter`) method and convenience function for `DataIoInterfaces`
The convenience function `load_dataset` doesn't provide a possibility to pass kwargs
https://github.com/glotaran/pyglotaran/blob/f214d78d6d68d624effad4f2bdaab41b573b866e/glotaran/io/register.py#L81
But even if it would users would have to look up the source code to see which possible options they have and what they are doing, since `help()` or IDE tools wouldn't work. Thus to properly load data a functionality that returns the method itself (with docstring and all) would be crucial.

@jsnel and @joernweissenborn I think the only controversial point was the splitting of `Io` so looking forward to your feedback 😄

s-weigand:
More a rewrite than a refactor but oh well 🤷‍♀️

Key features:
- All plugin functionality is in one package
- New plugins won't overwrite existing ones anymore
- It is now possible to pass additional keyword arguments to plugin function e.g. `load_dataset`
- Less copy pasta code due to abstraction
- Help functions for the underlying plugin implementations (useful when they have extra option)
- Overwriting of files is disabled by default, but can be explicitly allowed if needed (e.g. internal use or result overwriting)
- Added plugin support table, showing which plugin implemented which methods
- Added docs on how to write an Io plugin and moved quickstart its own subfolder 
- Fixed broken `AsciiDataIo`
- Moved save result to `YmlProjectIo` since besides saving the data to folder it tries to serialize the result to yaml (not fixed)
- Renamed `known_xy` to `is_known_xy`, to differentiate it more from `known_xys`
- Now all Io functions try to inferr the plugin to be used
- Renamed `read_xy` methods to `load_xy` that way they are consistent with the functions and there is less confusion
- Renamed `write_xy` methods and function to `save_xy` that way the naming pattern is consistent `load` / `save` (thx @jsnel for pointing that out 😄 )
- Added 102 new tests for the plugin system (100% tested)
- No Mypy type error (at least in the `plugin_system` itself `mypy glotaran/ | grep plugin_system`)
- 100% documented (passes darglint and pydocstyle)

Not sure what the best way to review it is, on the one hand each commit does more or less one thing on the other hand there is a lot of moving stuff around. 🤷‍♀️

Don't get scared by the 3k+ added lines, most of it are docs and only about 300 lines of code 😋 

If you want to try it out [here is the binder link to the plugin guide](https://mybinder.org/v2/gh/s-weigand/pyglotaran/refactor-plugin-system?urlpath=lab%2Ftree%2Fdocs%2Fsource%2Fnotebooks%2Fplugin_system%2Fplugin_howto_write_a_io_plugin.ipynb)

As a side note:
When writing the plugin guide I considered adding a `dev_mode` flag to the register decorators to avoid the plugin overwrite protection, since it would be easier to write a plugin in a notebook. But ultimately that would be dangerous if it was forgotten and a plugin method should be just a wrapper for the functionality + in the end the developing experience with a proper IDE is far better anyway.

**Testing**

**Closing issues**

- [x] Move all the plugin registration code in one module/package
- [x] Add warning when a plugin replaces an existing one

closes #586


s-weigand:
The doc-links don't pass, because I added a link that will only exist after merging.
As for the sonarcloud warings:
- [`"type" does not have a "__getitem__" method`](https://sonarcloud.io/project/issues?id=glotaran_pyglotaran&open=AXg8vkKycjCgjTHxoldH&pullRequest=600&resolved=false&types=BUG): pyflakes isn't up to all the new py310 syntax for tying so it doesn't know what `type[Model]` is
- [`he return value of "help" must be used`](https://sonarcloud.io/project/issues?id=glotaran_pyglotaran&open=AXg8vkKycjCgjTHxoldI&pullRequest=600&resolved=false&types=BUG) we could prevent by returning help ... which is still `None`, but for the tests it just doesn't get what the output is captured

jsnel (approved):
This PR represents a massive overhaul to the plugin system, with the addition of good documentation and a running notebook example on how to write your own. Granted, in some aspects it still a diamond in the rough, and could do with some polishing here and there but it's already more than good enough in the current state so I'm ok with merging this in. Future enhancements are planned, but out of scope for this initial rework/rewrite. 

Thank you @s-weigand for this tour de force.

PR #611, opened 2021-03-22
Make Models py310 ready

### LINKED ISSUE #610: Models will break in python 3.10
* glotaran version: all
* Python version: 3.10 (or before when using `from __future__ import annotations`)
* Operating System: all

### Description

The problem is that `Model` internally uses the `__annotations__` property
https://github.com/glotaran/pyglotaran/blob/a7bc0288526956c6f07f87d724231adb6a8317b9/glotaran/model/base_model.py#L44-L58
which is a `Mapping[str, Type]` if `from __future__ import annotations` isn't used.

But when `from __future__ import annotations` is used (default behavior in python 3.10), it becomes a `Mapping[str, str]` and we get this error all over the place:
```console
glotaran\model\base_model.py:58: in from_dict
    model_set(label, item_cls.from_dict(item))
E   AttributeError: 'str' object has no attribute 'from_dict'
```

Here is a little demo of the behavior change when using `from __future__ import annotations`.

```python
In [1]: from typing import List

In [2]: def foo(args=List[str]) -> List[str]:
   ...:     return args
   ...:

In [3]: foo.__annotations__
Out[3]: {'return': typing.List[str]}

In [4]: from __future__ import annotations

In [5]: def bar(args=List[str]) -> List[str]:
   ...:     return args
   ...:

In [6]: bar.__annotations__
Out[6]: {'return': 'List[str]'}
```

### What I Did

Add `from __future__ import annotations` at the top of `glotaran/model/decorator.py` and run `pytest`

s-weigand:
Instead of relying on the `__annotations__` property to be the old style (python < 3.10), `wrap_func_as_method` needs to set `__annotations__` explicitly where needed.

Since only `Model` accesses `__annotations__` directly to use them, I don't think we need the same precautions for `ModelProperty` and `model.attributes`, since they use `_glotaran_model_attribute_typed` and `_glotaran_properties` internally, which then is used by `Model`.

**Testing**

**Closing issues**

closes #610

### DIFF
--- a/glotaran/model/decorator.py
+++ b/glotaran/model/decorator.py
@@ -1,4 +1,5 @@
 """The model decorator."""
+from __future__ import annotations
 
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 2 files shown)


joernweissenborn (approved):

jsnel (dismissed):
👍

jsnel (approved):
👍

Issue #606, opened 2021-03-19
Megacomplex scale must be defined per dataset and per megacomplex


joernweissenborn:
Currently, a mega complex scale is applied to all dataset, but it must be per dataset

Issue #605, opened 2021-03-19
Combine KMatrices: change order of override


joernweissenborn:
Currently, if you have a lists of kmatrices in a megacomplex like

`k_matrix: [km1, km2]`

km1 would override km2 components which are present in both. it should be the other way around

PR #612, opened 2021-03-26
Fix several issues with KineticImage model

### LINKED ISSUE #604: normalize irf has no function
Currently, the property `normalize` can be se for multi-guassian irfs, but it is not used in the calculation.

joernweissenborn:
Closes #604, #605 and #606

### DIFF
--- a/glotaran/builtin/models/kinetic_image/irf.py
+++ b/glotaran/builtin/models/kinetic_image/irf.py
@@ -19,7 +19,7 @@ class IrfMeasured:
         "center": List[Parameter],
         "width": List[Parameter],
         "scale": {"type": List[Parameter], "allow_none": True},
-        "normalize": {"type": bool, "default": False},
+        "normalize": {"type": bool, "default": True},
         "backsweep": {"type": bool, "default": False},
         "backsweep_period": {"type": Parameter, "allow_none": True},
     },
--- a/glotaran/builtin/models/kinetic_image/k_matrix.py
+++ b/glotaran/builtin/models/kinetic_image/k_matrix.py
@@ -63,9 +63,9 @@ class KMatrix:
         """
         if not isinstance(k_matrix, KMatrix):
             raise TypeError("K-Matrices can only be combined with other K-Matrices.")
-        combined_matrix = {entry: k_matrix.matrix[entry] for entry in k_matrix.matrix}
-        for entry in self.matrix:
-            combined_matrix[entry] = self.matrix[entry]
+        combined_matrix = {entry: self.matrix[entry] for entry in self.matrix}
+        for entry in k_matrix.matrix:
+            combined_matrix[entry] = k_matrix.matrix[entry]
         combined = KMatrix()
         combined.label = f"{self.label}+{k_matrix.label}"
         combined.matrix = combined_matrix
--- a/glotaran/builtin/models/kinetic_image/k_matrix.py
+++ b/glotaran/builtin/models/kinetic_image/k_matrix.py
@@ -49,11 +49,13 @@ class KMatrix:
     def combine(self, k_matrix: KMatrix) -> KMatrix:
         """Creates a combined matrix.
 
+        When combining k-matrices km1 and km2 (km1.combine(km2)),
+        entries in km1 will be overwritten by corresponding entries in km2.
+
         Parameters
         ----------
         k_matrix :
             KMatrix to combine with.
-
 
         Returns
         -------
--- a/glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
+++ b/glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
@@ -37,9 +37,9 @@ class KineticImageDatasetDescriptor(DatasetDescriptor):
 
         return scales, matrices
 
-    def has_k_matrix(self):
-        _, matrices = self.get_megacomplex_k_matrices()
-        return len(matrices) != 0
+    def has_k_matrix(self) -> bool:
+        _, k_matrices = self.get_megacomplex_k_matrices()
+        return len(k_matrices) != 0
 
     def compartments(self):
         compartments = []
--- a/glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
+++ b/glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
@@ -16,11 +16,8 @@ class KineticImageDatasetDescriptor(DatasetDescriptor):
         scales = (
             [
                 self.megacomplex_scale[i]
-                for i in [
-                    i
-                    for i, megacomplex in enumerate(self.megacomplex)
-                    if megacomplex.has_k_matrix()
-                ]
+                for i, megacomplex in enumerate(self.megacomplex)
+                if megacomplex.has_k_matrix()
             ]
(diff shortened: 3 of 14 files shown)


glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
```diff
@@ -12,11 +12,38 @@
...
+                for megacomplex in self.megacomplex
+                if megacomplex.has_k_matrix()
+            ]
+        ]
+
+        return scales, matrices
+
+    def has_k_matrix(self):
+        _, matrices = self.get_megacomplex_k_matrices()
+        return len(matrices) != 0
```
s-weigand, line 42:
```suggestion
    def has_k_matrix(self) -> bool:
        _, k_matrices = self.get_megacomplex_k_matrices()
        return len(k_matrices) != 0
```
Just for concistency.

glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
```diff
@@ -12,11 +12,38 @@
...
+    def get_megacomplex_k_matrices(self):
+        scales = (
+            [
+                self.megacomplex_scale[i]
+                for i in [
+                    i
+                    for i, megacomplex in enumerate(self.megacomplex)
+                    if megacomplex.has_k_matrix()
+                ]
+            ]
```
s-weigand, line 24:
```suggestion
            [
                self.megacomplex_scale[i]
                    i
                    for i, megacomplex in enumerate(self.megacomplex)
                    if megacomplex.has_k_matrix()
            ]
```

This should do the same but with one less loop and without using `i` as a variable name twice.

Also since  we have two loops which check for `megacomplex.has_k_matrix()` we might want to create a filtered version and use that instead, e.g.:

```python
enumerated_megacomplices_with_k_matrix = [
    index, megacomplex
    for index, megacomplex in enumerate(self.megacomplex)
    if megacomplex.has_k_matrix()
]
```

glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
```diff
@@ -12,11 +12,38 @@
...
+        )
+
+        matrices = [
+            matrix
+            for matrix in [
+                megacomplex.full_k_matrix()
+                for megacomplex in self.megacomplex
+                if megacomplex.has_k_matrix()
+            ]
+        ]
```
s-weigand, line 36:
```suggestion
        matrices = [
                megacomplex.full_k_matrix()
                for megacomplex in self.megacomplex
                if megacomplex.has_k_matrix()
        ]
```
This should do the same but with one less loop.

s-weigand (commented):
I found some minor consistency and overcomplexity issues.
@jsnel needs to decide on the actual impact of the changes.

glotaran/builtin/models/kinetic_image/irf.py
```diff
@@ -19,7 +19,7 @@ class IrfMeasured:
         "center": List[Parameter],
         "width": List[Parameter],
         "scale": {"type": List[Parameter], "allow_none": True},
-        "normalize": {"type": bool, "default": False},
+        "normalize": {"type": bool, "default": True},
```
jsnel, line 22: Changing the default value for a model option is a breaking change, unless before this was not doing what it was supposed to, but even then it should be clearly communicated. 
This is something for the release notes @s-weigand .

joernweissenborn: THis not a change in default behaviour since the parameter was never used in the code. But everyone wants normalized irfs except a few cases, this is why I made it default `true`

glotaran/builtin/models/kinetic_image/kinetic_image_dataset_descriptor.py
joernweissenborn: This has_kmatrix will be obsolete in future since I will revice megacomplex/model composition.

glotaran/builtin/models/kinetic_image/kinetic_image_matrix.py
```diff
@@ -29,9 +29,9 @@ def kinetic_matrix(
         raise Exception(
             f'No initial concentration specified in dataset "{dataset_descriptor.label}"'
         )
-    initial_concentration = dataset_descriptor.initial_concentration.normalized(dataset_descriptor)
+    initial_concentration = dataset_descriptor.initial_concentration.normalized()
 
-    for k_matrix in k_matrices:
+    for i, k_matrix in enumerate(k_matrices):
```
s-weigand, line 34:
```suggestion
    for k_matrix_index , k_matrix in enumerate(k_matrices):
```
That LGTM won't complain about the use of `i` in nested for loops and it is more apparent which `i` does what.

glotaran/builtin/models/kinetic_image/kinetic_image_matrix.py
```diff
@@ -46,6 +46,9 @@ def kinetic_matrix(
             matrix_implementation,
         )
 
+        if megacomplex_scales is not None:
+            this_matrix *= megacomplex_scales[i]
```
s-weigand, line 50:
```suggestion
            this_matrix *= megacomplex_scales[k_matrix_index]
```
That LGTM won't complain about the use of `i` in nested for loops and it is more apparent which `i` does what.

s-weigand (commented):
Found another small issue, LGTM was justifiedly complaining about.

glotaran/builtin/models/kinetic_image/k_matrix.py
```diff
@@ -63,9 +63,9 @@ def combine(self, k_matrix: KMatrix) -> KMatrix:
         """
```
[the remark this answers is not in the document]
jsnel: Added to docstring

glotaran/builtin/models/kinetic_image/kinetic_image_matrix.py
jsnel: Added in commit "Avoid using same variable name in nested for loops"

glotaran/builtin/models/kinetic_image/kinetic_image_matrix.py
jsnel: Added in commit "Avoid using same variable name in nested for loops"

jsnel (dismissed):
LGTM :sparkles:

glotaran/builtin/models/kinetic_image/k_matrix.py
```diff
@@ -48,13 +48,14 @@ def involved_compartments(self) -> list[str]:
 
     def combine(self, k_matrix: KMatrix) -> KMatrix:
         """Creates a combined matrix.
+        When combining k-matrices a and b (a.combine(b)),
+        entries in a will overwrite by corresponding entries in b.
```
s-weigand, line 52:
```suggestion

        When combining k-matrices a and b (a.combine(b)),
        entries in a will overwrite by corresponding entries in b.
```
Sorry for being nitpicky, but this would be one less error to fix when we at some point are able to add pydocstyle.

s-weigand (approved):
Besides a nitpicky thing with a docstring LGTM

jsnel (approved):
LGTM ✨

PR #620, opened 2021-03-29
✨ Add Folder plugin for save result

### LINKED ISSUE #619: save_result needs a folder plugin
pyglotaran: main branch (v0.4-to_be) 

### Description

It would be nice if this would 'just work':
```python
save_result(result_path=str(output_folder), result=result, allow_overwrite=True)
# or if format must be given:
save_result(result_path=str(output_folder), result=result, format_name="folder", allow_overwrite=True)
```
In the meantime, a fallback is
```python
save_result(result_path=str(output_folder), result=result, format_name="yml", allow_overwrite=True)
# but make sure the output_folder is empty or you get an error that files exist despite allow_overwrite=True
```

s-weigand:
This plugin adds back the functionality from `glotaran.analysis.result.Result.save(result_path)` of pyglotaran `0.3.x`.
For `glotaran.io.save_result` it also adds inferring of `format_name=="folder"` if the provided `result_path` doesn't have an extension.
If `result_path` is a file it will raise a `ValueError`.

Usage equivalent to `result.save(result_path)`
```python
save_result(result_path=str(output_folder), result=result, allow_overwrite=True)
# or if format must be given:
save_result(result_path=str(output_folder), result=result, format_name="folder", allow_overwrite=True)
save_result(result_path=str(output_folder), result=result, format_name="pyglotaran-legacy", allow_overwrite=True)
```

Also, the plugin is registered under the names `folder` and `pyglotaran-legacy`, thus if we want to fundamentally change the `folder`-plugin we can keep the old implementation around as `pyglotaran-legacy` for backward compatibility.

**Testing**

**Closing issues**

closes #619

### DIFF
--- a/glotaran/builtin/io/folder/__init__.py
+++ b/glotaran/builtin/io/folder/__init__.py
@@ -1,0 +1,1 @@
+"""Plugin to dump pyglotaran object as files in a folder."""
--- a/glotaran/builtin/io/folder/folder_plugin.py
+++ b/glotaran/builtin/io/folder/folder_plugin.py
@@ -1,0 +1,74 @@
+"""Implementation of the folder Io plugin.
+
+The current implementation is an exact copy of how ``Result.save(path)``
+worked in glotaran 0.3.x and meant as an compatibility function.
+"""
+
+from __future__ import annotations
+
+import os
+from typing import TYPE_CHECKING
+
+from glotaran.io.interface import ProjectIoInterface
+from glotaran.plugin_system.project_io_registration import register_project_io
+
+if TYPE_CHECKING:
+    from glotaran.project import Result
+
+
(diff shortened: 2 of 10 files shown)


jsnel (approved):
Tested OK.

jsnel (changes requested):
Please change 'pyglotaran-legacy' to 'legacy' in the context of pyglotaran plugins that should be sufficient.

s-weigand:
Admittedly the importing of the fixture and that for added `noqa`s look a bit ugly, but since the tests are inside of the package and not in a separate folder we can just use a global `conftest.py` so all fixtures are known to all pytest tests.

jsnel (approved):
LGTM.  More✨ for pyglotaran

PR #630, opened 2021-04-03
Avoid crash in result markdown printing when optimization fails

### LINKED ISSUE #629: Crash when optimization fails (lm_result is None)
* glotaran version: main

### Description

Optimize a model that happens to fail the optimization, and return `lm_result=None`
a type error occurs when printing the Markdown results and results (from before the termination of the optimizer are not plotted).

### What I Did

```
Executing: ex_four_datasets_two_guidance_no_area_irf.py 

Your model is valid.
   Iteration     Total nfev        Cost      Cost reduction    Step norm     Optimality

glotaran\analysis\optimize.py:62: UserWarning: Optimization failed:
zero-size array to reduction operation maximum which has no identity
  warn(f"Optimization failed:\n\n{e}")
Traceback (most recent call last):
  File "ex_four_datasets_two_guidance_no_area_irf.py", line 98, in <module>
    main()
  File "ex_four_datasets_two_guidance_no_area_irf.py", line 64, in main
    print(result.markdown(True))
  File "glotaran\project\result.py", line 126, in markdown
    ["Chi Square", f"{self.chi_square:.2e}"],
TypeError: unsupported format string passed to NoneType.__format__
```

jsnel:
Avoid crash in Result's print markdown method

**Closing issues**

Closes #629

### DIFF
--- a/glotaran/project/result.py
+++ b/glotaran/project/result.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from math import nan
 
 import numpy as np
 import xarray as xr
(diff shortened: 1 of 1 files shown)


glotaran/project/result.py
```diff
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from math import nan
```
s-weigand, line 5: Why use another import when we already have `numpy` and can use `np.nan`?

s-weigand (approved):
LGTM

glotaran/project/result.py
jsnel: Changed PR to avoid additional import

PR #631, opened 2021-04-03
🩹 Add deprecation framework and readd 0.3.x API as deprecated

### LINKED ISSUE #588: How to handle breaking API changes? (#556 follow up)
We recently had some breaking API changes in #556, BUT the package `is an early access release` so "the right way to do it", keeping the old API alive and warning the users that it will be dropped in the future might slow down the development (at least I was told so).
> Warning: This is an early access release, please refer to the usage notice down below prior to committing to use pyglotaran to avoid surprises down the line.

IMHO there are only 2 user-friendly ways and don't forget that we are our own users (e.g. atm we have at least 2 repos that depend on the core  [pyglotaran_extras](https://github.com/glotaran/pyglotaran_extras) and [pyglotaran_examples](https://github.com/glotaran/pyglotaran_examples))
1.) Do it "the right way" and redirect the old API calls to the new ones, warning users.
2.) Use a "dummy API" which breaks, but tells the users what the new way is.

I'm making this an issue now because IMHO this is a workflow pattern that will influence future development and having a set way to do things frees us from the mental load to think about it each time (e.g. think back of the times before we had `pre-commit` doing all the formatting and lining for us). 
Also, a more graceful way to handle breaking changes to the API profits code in our own organization.

s-weigand:
This PR adds a deprecation framework and reintroduces the API of glotaran 0.3.x giving deprecation warnings when used, which instruct users how to change their code.

As usual with my PRs you best review it commit by commit, since they are structured by responsibility.
It might be best to start with the updated contributing docs where I describe the intended usage and then review the rest in order.

**Testing**

**Closing issues**

closes #588

### DIFF
--- a/glotaran/deprecation/deprecation_utils.py
+++ b/glotaran/deprecation/deprecation_utils.py
@@ -38,7 +38,7 @@ class OverDueDeprecation(Exception):
 
 
 def glotaran_version() -> str:
-    """Pyglotaran version of the distribution.
+    """Version of the distribution.
 
     This is basically the same as ``glotaran.__version__`` but independent from glotaran.
     This way all of the deprecation functionality can be used even in
@@ -48,7 +48,7 @@ def glotaran_version() -> str:
     Returns
     -------
     str
-        Versionstring of glotaran.
+        The version string.
     """
     return distribution("pyglotaran").version
 
--- a/glotaran/deprecation/deprecation_utils.py
+++ b/glotaran/deprecation/deprecation_utils.py
@@ -198,7 +198,10 @@ def warn_deprecated(
             return load_model(model_path)
 
     """
-    if parse_version(glotaran_version()) >= parse_version(to_be_removed_in_version):
+    if (
+        parse_version(glotaran_version()) >= parse_version(to_be_removed_in_version)
+        and "dev" not in glotaran_version()
+    ):
         raise OverDueDeprecation(
             f"Support for {deprecated_qual_name_usage.partition('(')[0]!r} was "
             f"supposed to be dropped in version: {to_be_removed_in_version!r}\n"
--- a/glotaran/deprecation/deprecation_utils.py
+++ b/glotaran/deprecation/deprecation_utils.py
@@ -1,0 +1,504 @@
+"""Helper functions to give deprecation warnings."""
+
+from __future__ import annotations
+
+import os
+import sys
+from functools import wraps
+from importlib import import_module
+from importlib.metadata import distribution
+from types import ModuleType
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Callable
+from typing import TypeVar
+from typing import cast
+from warnings import warn
+
+DecoratedCallable = TypeVar(
+    "DecoratedCallable", bound=Callable[..., Any]
+)  # decorated function or class
+
+if TYPE_CHECKING:
+    from typing import Sequence
+
+
+class OverDueDeprecation(Exception):
+    """Error thrown when a deprecation should have been removed.
+
+    See Also
+    --------
+    deprecate
+    warn_deprecated
+    deprecate_module_attribute
+    deprecate_submodule
+    """
+
+    pass
+
+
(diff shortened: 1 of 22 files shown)


glotaran/deprecation/modules/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Package containing deprated code implementation which were removed.
```
jsnel, line 1: deprated -> deprecated 
(perhaps remove 'code')
implementation -> implementations

glotaran/deprecation/deprecation_utils.py
```diff
@@ -0,0 +1,504 @@
...
+    warn_deprecated
+    deprecate_module_attribute
+    deprecate_submodule
+    """
+
+    pass
+
+
+def glotaran_version() -> str:
+    """Pyglotaran version of the distribution.
```
jsnel, line 41: pyglotaran with a capital here?
"Version of the distribution" is probably enough.

glotaran/deprecation/deprecation_utils.py
```diff
@@ -0,0 +1,504 @@
...
+
+    Returns
+    -------
+    str
+        Versionstring of glotaran.
+    """
+    return distribution("pyglotaran").version
+
+
+def parse_version(version_str: str) -> tuple[int, int, int]:
```
jsnel, line 56: This doesn't seem to take the -qualifier into account (e.g. -dev -develop -development). 
What if we release 0.5.0 and now change the version sources to 0.6.0-develop … will the stuff that was marked for deprecation now immediately break? or is -dev still allowed?
0.6.0-develop should be treated as 0.5.x+1 (where x is the last minor released version) or if deprecation can only happen in minor version bumps anyways, 0.6.0-develop is just 0.5.0 as far as the deprecation API is concerned

glotaran/deprecation/deprecation_utils.py
```diff
@@ -0,0 +1,504 @@
...
+
+    See Also
+    --------
+    deprecate
+    warn_deprecated
+    deprecate_module_attribute
+    deprecate_submodule
+    """
+
+    pass
```
jsnel, line 37: Is this pass here needed? codacy [doesn't think so](https://app.codacy.com/gh/glotaran/pyglotaran/pullRequest?prid=7203405).

glotaran/deprecation/deprecation_utils.py
```diff
@@ -0,0 +1,504 @@
...
+
+    This is basically the same as ``glotaran.__version__`` but independent from glotaran.
+    This way all of the deprecation functionality can be used even in
+    ``glotaran.__init__.py`` without moving the import below the definition of
+    ``__version__`` or causeing a circular import issue.
+
+    Returns
+    -------
+    str
+        Versionstring of glotaran.
```
jsnel, line 51: "Versionstring of glotaran." -> "The version string"

glotaran/deprecation/modules/glotaran_root.py
```diff
@@ -0,0 +1,139 @@
+"""Deprecated attributes from ``glotaran.__init__`` which ere removed."""
```
jsnel, line 1: ere -> are

jsnel (changes requested):
Overall ok, just a request to handle the deprecation of development versions differently.

glotaran/deprecation/deprecation_utils.py
jsnel: Addressed in [this commit ](https://github.com/glotaran/pyglotaran/pull/631/commits/1c5ded3912463db868521f8d39aaf5d1daf42945)

glotaran/deprecation/deprecation_utils.py
s-weigand: Ohhh nice this was new to me, looks like when you define a docstring you can skip the `pass`, which else is required.

jsnel (approved):
LGTM 🚀

PR #647, opened 2021-04-22
🩹 Fix exception in sdt reader index calculation

### LINKED ISSUE #646: 🐛 Sdt reader index calculation is broken
* glotaran version: main

### Description

When reading sdt files w/o providing an index the reader raises an exception when calculating the `index`.

https://github.com/glotaran/pyglotaran/blob/f4e264512ffdb4c0ce98956719bd1bd165e77e87/glotaran/builtin/io/sdt/sdt_file_reader.py#L112

Since `data` at this point in the code is an `np.ndarray` with shape `(spectral_dimension, temporal_dimension)` the index can't be calculated this way.

### What I Did

I tried to read the sdt files from FJ and `load_dataset`raised an exception.

s-weigand:
- index is calculated correctly + test
- differentiate between raw_data (np.ndarray) and data (xr.Dataset|xr.DataArray)
- warning raised at the correct stacklevel

**Testing**

**Closing issues**

closes #646

### DIFF
--- a/glotaran/builtin/io/sdt/sdt_file_reader.py
+++ b/glotaran/builtin/io/sdt/sdt_file_reader.py
@@ -73,15 +73,16 @@ class SdtDataIo(DataIoInterface):
                         f"By default only the first Dataset will be read. "
                         f"If you only need the first Dataset and want get rid of "
                         f"this warning you can set dataset_index=0."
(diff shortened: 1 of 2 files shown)


jsnel (approved):
LGTM

PR #655, opened 2021-05-16
Update codeowners validator gh-action

jsnel:
Trying improvement found in `remal`'s `oss-template`

**Testing**

This should not error on the codeowners validator check.


jsnel:
Fixed in #656

PR #656, opened 2021-05-17
🔧 Change Codeowners Validator workflow to use pull_request_target

### LINKED ISSUE #506: Improvement to CODEOWNERS sanity workflow
The sanity workflow can be triggered manually using workflow_dispatch but as per suggestion of @s-weigand "it might also be good to run it on `push` and/or `pull_request` in case the `CODEOWNERS` file gets changed e.g. if we add a new subpackage with specific owners.
Since the workflow isn't coupled to other steps and should be pretty quick this shouldn't produce a big overhead."

_Originally posted by @s-weigand in https://github.com/glotaran/pyglotaran/pull/500#discussion_r559195208_

s-weigand:
Changed triggers for codeowners validation workflow:
- PRs use `pull_request_target` and explicit checkout
- `push` only triggers on `main`
- Dropped `schedule` since it runs on pushes to `main`

**Testing**

**Closing issues**

closes #506

### DIFF
--- a/.github/workflows/codeowners_validator.yml
+++ b/.github/workflows/codeowners_validator.yml
@@ -2,18 +2,28 @@ name: "Codeowners Validator"
 
 on:
   push:
-  schedule:
-    # Runs at 08:00 UTC every day
-    - cron:  '0 8 * * *'
+    branches:
+      - "main"
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


jsnel (approved):
🚀

PR #663, opened 2021-05-21
🔧 Removed custom script in favour of nbstripout --strip-empty-cells

s-weigand:
`nbstripout==0.4.0` added the option to remove empty cells from notebooks (`--strip-empty-cells`) so we can drop my lil hacky script to do that and use this feature instead 😄 

**Testing**


jsnel (approved):
Nice, peel of that :adhesive_bandage: and show the ✨

jsnel (approved):
Nice, peel of that :adhesive_bandage: and reveal the :sparkles:

PR #665, opened 2021-05-21
♻️ Polish the plugin system

### LINKED ISSUE #664: ♻️ Polish the plugin system
* glotaran version: 0.4.0-dev

### Description

The new plugin system still has some kinds, which should be polished out:
- `save` functions of IO plugins have an uncommon argument order of (`<path>`, `<object>`) while python uses (`<object>`, `<path>`) (e.g. [json.dump](https://docs.python.org/3/library/json.html#basic-usage))
- register functions give a warning when a plugin overwrites itself, which can be annoying when developing a new plugin in a notebook (see [demo guide](https://mybinder.org/v2/gh/glotaran/pyglotaran/f366ed6?urlpath=lab/tree/docs/source/notebooks/plugin_system/plugin_howto_write_a_io_plugin.ipynb))
- While the plugin name extending on conflicts prevents plugins from disappearing, the fact that this only happens on conflicts can make the analysis not reproducible depending on which plugins are installed
- `save` and `load` IO function only accepts strings, so users need to convert paths manually, for convenience this can be done inside the plugin wrapper functions
- `PluginOverwriteWarning`s point to [`add_plugin_to_registry`](https://github.com/glotaran/pyglotaran/blob/f366ed6609b1276b705c34c84a935cc0d40cf8c6/glotaran/plugin_system/base_registry.py#L175) rather than the plugin which isn't very helpful for users

s-weigand:
This PR removes the issues I found with the new plugin system.

- `save` functions for IO plugins now have a natural python argument order ( (`<object>`, `<path>`) instead of (`<path>`, `<object>`))
- register functions don't trow a `PluginOverwriteWarning` if the plugin overwrites itself
- plugins now register at least twice, once as the developer intended and once under their full name (import path+instance postfix if it is an instantiated plugin like the IO plugins)
- new `set_plugin` functionality allowing to pin plugins and thus allowing reproducible analysis no matter which plugins have conflicts
- `project_io_plugin_table` and `data_io_plugin_table` now also show the plugin names if wanted, which helps to identify which plugin is used for which format
- `save` and `load` IO functions now also accept `PathLike` paths and convert them to strings before passing them to the plugins
- `PluginOverwriteWarning`s now point to the plugin trying to register

Since we didn't have a release with the new plugin system yet, IMHO there shouldn't be a need for proper deprecation of changed argument order of the `save` functions.

**Testing**

**Closing issues**

closes #664

### DIFF
--- a/glotaran/plugin_system/data_io_registration.py
+++ b/glotaran/plugin_system/data_io_registration.py
@@ -106,7 +106,7 @@ def known_data_formats(full_names: bool = False) -> list[str]:
 
     Parameters
     ----------
-    full_names: bool
+    full_names : bool
         Whether to display the full names the plugins are
         registered under as well.
 
@@ -135,7 +135,7 @@ def set_data_plugin(
     Parameters
     ----------
     format_name : str
-        Format to use the plugin for.
+        Format name used to refer to the plugin when used for ``save`` and ``load`` functions.
     full_plugin_name : str
         Full name (import path) of the registered plugin.
     """
@@ -180,7 +180,7 @@ def load_dataset(
         File containing the data.
     format_name : str
         Format the file is in, if not provided it will be inferred from the file extension.
-    **kwargs: Any
+    **kwargs : Any
         Additional keyword arguments passes to the ``read_dataset`` implementation
         of the data io plugin. If you aren't sure about those use ``get_dataloader``
         to get the implementation with the proper help and autocomplete.
@@ -207,7 +207,7 @@ def save_dataset(
 
     Parameters
     ----------
-    dataset: xr.Dataset|xr.DataArray
+    dataset : xr.Dataset | xr.DataArray
         Data to be written to file.
     file_name : str | PathLike[str]
         File to write the data to.
@@ -215,7 +215,7 @@ def save_dataset(
         Format the file should be in, if not provided it will be inferred from the file extension.
     allow_overwrite : bool
         Whether or not to allow overwriting existing files, by default False
-    **kwargs: Any
+    **kwargs : Any
         Additional keyword arguments passes to the ``write_dataset`` implementation
         of the data io plugin. If you aren't sure about those use ``get_datawriter``
         to get the implementation with the proper help and autocomplete.
@@ -293,9 +293,9 @@ def data_io_plugin_table(*, plugin_names: bool = False, full_names: bool = False
 
     Parameters
     ----------
-    plugin_names:bool
+    plugin_names : bool
         Whether or not to add the names of the plugins to the table.
-    full_names: bool
+    full_names : bool
         Whether to display the full names the plugins are
         registered under as well.
 
--- a/glotaran/plugin_system/data_io_registration.py
+++ b/glotaran/plugin_system/data_io_registration.py
@@ -75,6 +75,7 @@ def register_data_io(
             plugin_register_keys=format_names,
             plugin_class=cls,
             plugin_registry=__PluginRegistry.data_io,
+            plugin_set_func_name="set_data_plugin",
         )
         return cls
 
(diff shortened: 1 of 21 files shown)


glotaran/plugin_system/data_io_registration.py
```diff
@@ -98,15 +101,49 @@ def is_known_data_format(format_name: str) -> bool:
...
+
+    Effected functions:
+
+    - :func:`load_dataset`
+    - :func:`save_dataset`
+
+    Parameters
+    ----------
+    format_name : str
+        Format to use the plugin for.
```
jsnel, line 138: A bit vague what a "format" is. Perhaps refer to the documentation where it is explained in more detail.
In most cases it would refer to a file extension, but what about (future) support for data folders, e.g. a folder containing a specific set of files that together describe a dataset? Doesn't have to be addressed in this PR.

glotaran/plugin_system/data_io_registration.py
```diff
@@ -170,12 +207,12 @@ def save_dataset(
 
     Parameters
     ----------
-    file_name : str
+    dataset: xr.Dataset|xr.DataArray
```
jsnel, line 210: spaces around | ?

glotaran/plugin_system/data_io_registration.py
```diff
@@ -249,20 +286,35 @@ def show_data_io_method_help(
...
+def data_io_plugin_table(*, plugin_names: bool = False, full_names: bool = False) -> str:
     """Return registered data io plugins and which functions they support as markdown table.
 
     This is especially useful when you work with new plugins.
 
+    Parameters
+    ----------
+    plugin_names:bool
+        Whether or not to add the names of the plugins to the table.
+    full_names: bool
```
jsnel, line 298: fyi, you mix and match
a: b
and
a : b

glotaran/plugin_system/project_io_registration.py
```diff
@@ -423,20 +470,35 @@ def show_project_io_method_help(format_name: str, method_name: ProjectIoMethods)
...
 
-def project_io_plugin_table() -> str:
+def project_io_plugin_table(*, plugin_names: bool = False, full_names: bool = False) -> str:
     """Return registered project io plugins and which functions they support as markdown table.
 
     This is especially useful when you work with new plugins.
 
+    Parameters
+    ----------
+    plugin_names:bool
```
jsnel, line 480: space : space

jsnel (changes requested):
A comment (to think about) and some suggestion to correct formatting in docstrings (so we don't have to touch it in a later commit)

glotaran/plugin_system/data_io_registration.py
s-weigand: Yeah saw that now that you pointed it out 😅 
`a : b` comes from the autogenerated docstring by from the VSCode extension [Python Docstring Generator
](https://github.com/NilsJPWerner/autoDocstring)
`a: b` Is from copy-pasting out of the signature
`a:b` Is from copy-pasting out of the signature, before saving the file an black properly formatting it

I want an autformatter for docstrings! 😆

glotaran/plugin_system/data_io_registration.py
s-weigand: How about "Format name to be used with ``save`` and ``load`` functions."? 
Since the argument has the same name as the one in `load_dataset` and `save_dataset`

s-weigand: As for a data folder plugin, you could just register it under the name `data_folder` and raise an error in the `save` and `load` methods if `file_name` isn't a folder. The only restriction we have on format names is that they aren't allowed to contain `.`, since this is the quick and dirty way to differentiate between full names and intended names.

jsnel:
```rst
    Parameters
    ----------
    format_name : str
        Format name used to refer to the plugin when used for ``save`` and ``load`` functions.
```

jsnel (approved):
Thanks!

PR #671, opened 2021-05-27
Feature: Multiple Megacomplex Types Per Model

joernweissenborn:
This PR adds the possibility of defining multiple megacomplex types per model. It also reliefs models of needing to implement megacomplex combination.

Key changes
------------------

* Models can now define multiple megacomplex types
* Models have a default megacomplex type, which is implicit derived if not specified
* If the type of a megacomplex is not specified in a model spec, the default megacomplex type of the model is used
* The `model_matrix` function is gone. Instead of a matrix function per model, the megacomplexes defined by the model must implement the `calculate_matrix` function
* The actual matrices are not calculated on a dataset level, but on a megacomplex level. Megacomplex combination is now responsability of the `Problem` class.
* Baseline is now a megacomplex `KineticBaslineMegaComplex` instead of a dataset property
* Coherent Artifact is now a megacomplex `CoherentArtifactMegacomplex` instead of an Irf type
* `model.simulate` has been removed to avoid zircular dependencys

Deprecations
------------------
* megacomplex_types must be now children of `model.Megacomplex`
* `model.simulate` has been removed. `simulate` must be imported from analysis
* when specifying a megacomplex in list notation, the megacomplex type MUST be specified
* existing models using baselines and/or coherent artifacts must be adapted

Not Implemented in this PR
------------------------------------
* megacomplexes for different dimensions
* replace global matrix with megacomplexes for simulations


jsnel (approved):
Peer reviewed by @jsnel and @s-weigand 

Looks good to us.

PR #673, opened 2021-05-28
Add Channel/Wavelength dependend shift paramter to irf.

joernweissenborn:
This PR adds the possibility to model a per wavelength shift of a MultiGussianIrf.


jsnel (approved):
LGTM.

Note, coherent artifact IRF was removed in this PR since it was replaced with a megacomplex version in the PR #671

PR #676, opened 2021-05-29
♻️ Refactor IrfMultiGaussian.calculate to be more pythonic

s-weigand:
Just something we overlooked in the review of #673

**Testing**


jsnel (approved):
LGTM

PR #678, opened 2021-05-29
👌 Ensure Parameter.label is str or None

### LINKED ISSUE #634: Make ParameterGroup labels consistently str
* glotaran version: all
* Python version: all
* Operating System: all

### Description

Right now if you define parameters as a list, e.g.:
```yaml
foo:
  - 123
```
If you create a `ParameterGroup` from it e.g. with 

```python
parameters = load_parameters("foo:\n  - 123", format_name="yml_str")
```
Then you have to access it with an int (e.g. in a test)
```python
assert parameters["foo"].get(1) == 123
```

Todays decision by @joernweissenborn , @jsnel  and @s-weigand  was that it will be best to make consistently a `str`.

### Side node: 
`ParameterGroup.has(label)` also doesn't work as expected, in general, `ParameterGroup` is undertested.

s-weigand:
- Ensure that Paremeter.label always is a str or None
- Improve Parameter typing (no more mypy error in parameter.py)
- Fix error message on invalid group label
- Improved test coverage
- Improved docstrings

Typing error reduction from [268](https://github.com/glotaran/pyglotaran/pull/444#issuecomment-850829451) to [253](https://github.com/s-weigand/pyglotaran/runs/2700885616?check_suite_focus=true)

**Testing**

**Closing issues**

closes #634

### DIFF
--- a/glotaran/deprecation/modules/test/test_glotaran_root.py
+++ b/glotaran/deprecation/modules/test/test_glotaran_root.py
@@ -98,7 +98,7 @@ def test_read_parameters_from_yaml():
 
     assert isinstance(result, ParameterGroup)
     assert isinstance(result["foo"], ParameterGroup)
-    assert result["foo"].get(1) == 123
+    assert result["foo"].get("1") == 123
 
 
 def test_read_parameters_from_yaml_file(tmp_path: Path):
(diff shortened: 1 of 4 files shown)


jsnel (approved):
L👌TM

s-weigand:
As @joernweissenborn said there are still some artifacts in the parameter which originate from the `lmfit` time but we can clean those up later since they need some test refactoring aswell.

Just to document them:
- `Parameter.full_label`
- `Parameter.set_from_group`

Are considered artifacts, by @joernweissenborn 
> `set_from_group` and `full_label` are both `lmfit` artifacts

joernweissenborn (approved):

jsnel (approved):
LGTM

PR #679, opened 2021-05-29
🧪 Test depreciation of Model.simulate

s-weigand:
Follow up on #671 
This test is mainly used for the tests to fail when we make a new release and the deprecation is overdue.

**Testing**


jsnel (approved):
LGTM

PR #680, opened 2021-05-29
🔧 Remove the schedule from autoupdate-pre-commit-config

s-weigand:
pre-commit-ci will do the scheduled autoupdates from now on.
But in case we manually want to trigger it we still got this workflow to start an update PR from the actions tab on github.

**Testing**


jsnel (approved):
-⏲️

PR #684, opened 2021-06-04
🩹 Fix spectral-model crashing on SVD calculation

### LINKED ISSUE #683: 🐛 Problem._prepare_data crashes when used with spectral-model
* glotaran version: 0.4.0-dev0

### Description

When using a `spectral-model` `Problem._prepare_data` raises a `ValueError`

E.g.:
```
ValueError: conflicting sizes for dimension 'time': length 41 on 'data_left_singular_vectors' and length 81 on 'time'
```

Since `model` and `global` dimensions are transposed before calculating the `SVD`, the `SVD` is effectively transposed as well, so the dimensions don't match.

s-weigand:
Since the SVD is independent of the model, we can calculate it before transposing `model_dimension` and `global_dimension`.
I also refactored `glotaran.io.prepare_dataset` to reduce code duplication and prevent recalculation of the SVD, since data readers like `ascii` already calculate the SVD when reading the dataset.

**Testing**

**Closing issues**

closes #683

### DIFF
--- a/glotaran/analysis/problem.py
+++ b/glotaran/analysis/problem.py
@@ -11,6 +11,7 @@ import xarray as xr
 from glotaran.analysis.nnls import residual_nnls
 from glotaran.analysis.util import get_min_max_from_interval
 from glotaran.analysis.variable_projection import residual_variable_projection
+from glotaran.io.prepare_dataset import add_svd_to_dataset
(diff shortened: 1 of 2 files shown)


s-weigand:
Converted the PR to a draft since the residual SVD is transposed.

jsnel (approved):
Look good

Issue #674, opened 2021-05-28
Add spectral model to pyglotaran


jsnel:
### Description

Spectral model

PR #686, opened 2021-06-05
🩹 Fix calculation of skewed-gaussian in spectral model

### LINKED ISSUE #685: 🐛Calculation of 'lorentzian-bandshape' is wrong
* glotaran version: 0.4.0.-dev0

### Description
The calculation of `lorentzian-bandshape` uses the `nm` axis for calculations with is reciprocal to the energy.
https://github.com/glotaran/pyglotaran/blob/1de783ffaba9c3e45cfb162a01f3f9b000afe05f/glotaran/builtin/models/spectral/shape.py#L41-L52
This adds an extra deformation to the shape, which results in a bad residual  (dummy case study with data simulated by TIMP).

grafik

We should use `1/cm` like (TIMP) internally.

Also, it doesn't protect against `ZeroDivisionError`

### Expected risidual
grafik

s-weigand:
This PR changes the implementation of the ~~lorentzian-bandshape~~ skewed-gaussian calculation to use a `1/cm` instead of an `nm` axis.

Thus the `location` and `width` now also need to be provided in `1/cm`.

Also, this adds [proper docstrings with the equations.

**Testing**

**Closing issues**

closes #685 
completes implementation of spectral model #674

### DIFF
--- a/glotaran/analysis/simulation.py
+++ b/glotaran/analysis/simulation.py
@@ -55,10 +55,10 @@ def simulate(
     model_dimension = axes[model.model_dimension]
     global_dimension = axes[model.global_dimension]
 
-    dim1 = model_dimension.size
+    # dim1 = model_dimension.size
     dim2 = global_dimension.size
     result = xr.DataArray(
-        np.empty((dim1, dim2), dtype=np.float64),
+        data=0.0,
         coords=[
             (model.model_dimension, model_dimension),
             (model.global_dimension, global_dimension),
--- a/glotaran/builtin/models/kinetic_image/kinetic_image_result.py
+++ b/glotaran/builtin/models/kinetic_image/kinetic_image_result.py
@@ -37,7 +37,7 @@ def retrieve_species_associated_data(model, dataset, dataset_descriptor, name):
             model.global_dimension,
             "species",
         ),
-        dataset.clp.sel(clp_label=compartments),
+        dataset.clp.sel(clp_label=compartments).data,
     )
 
     if len(dataset.matrix.shape) == 3:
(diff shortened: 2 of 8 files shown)


s-weigand:
Maybe we should also rename `location` to `center`, so it is consistent with the naming for Gaussians (e.g. used for the IRF).

ism200:
Dear all,

With a skewed Gaussian (the spectral shape that is used most often) it is better to call it the location (of the maximum) parameter.

Regards,

ivo

Van: Sebastian Weigand ***@***.***>
Verzonden: zondag 6 juni 2021 00:29
Aan: glotaran/pyglotaran ***@***.***>
CC: Stokkum, I.H.M. van ***@***.***>; Review requested ***@***.***>
Onderwerp: Re: [glotaran/pyglotaran] 🩹 Fix calculation of 'lorentzian-bandshape' (#686)


Maybe we should also rename location to center, so it is consistent with the naming for Gaussians (e.g. used for the IRF).

jsnel (changes requested):
Given the current implementation follows the (skewed/unskewed) *Gaussian* implementation of TIMP (DOI: [10.18637/jss.v018.i03](http://dx.doi.org/10.18637/jss.v018.i03) page 29, 31-36, it would make more sense to rename the class accordingly. 
The TIMP paper, page 29, hints at possible other implementations:

> The spectral bandshapes are typically described in terms of a linear superposition of standard band shapes (e.g., Gaussian, Lorentzian, Voigt, skewed Gaussian) or in terms of splines.

jsnel (approved):
Renamed lorentzian to (skewed) gaussian
Addressed a (major) deprecation warning in xarray which would have resulted in an error in 0.19.0

Issue #687, opened 2021-06-06
👌 Better Notebook integration


s-weigand:
* glotaran version: all

## Automatic rendering of markdown

Right now at the beginning of each notebook I have to define helper functions to render markdown.

https://github.com/glotaran/pyglotaran/blob/1de783ffaba9c3e45cfb162a01f3f9b000afe05f/docs/source/notebooks/quickstart/quickstart.ipynb#L19-L31

This can be prevented by adding the [special method `_repr_markdown_`](https://ipython.readthedocs.io/en/latest/config/integrating.html?highlight=_repr_markdown_#rich-display) so just having `model` (or any other object that supports markdown) in the last line of your cell renders it properly as markdown.

## Adjustable heading level for model

`jupyterlab>=3` has a builtin TOC (Table Of Content) which allows you to quickly navigate your notebook.
Since the heading level of `Model` in the markdown representation of model starts with `h1`, this messes up the TOC.
![grafik](https://user-images.githubusercontent.com/9513634/120925130-6b930e80-c6d7-11eb-999b-8b40ae832cd6.png)

Wished output (model starts as `h3`)
![grafik](https://user-images.githubusercontent.com/9513634/120925157-8ebdbe00-c6d7-11eb-80a6-df4a027854d1.png)

We can easily solve this problem by adding a `base_heading_level` parameter to the `markdown` method.

## No markdown in `__repr__`
Not directly a notebook issue, but has to do with markdown.
Overwriting the magic method `__repr__` to use markdown leads to not helpful and cluttered test assertions/tracebacks.

Consider the following fake test file (`dummy_test.py`):
```python
from glotaran.parameter import ParameterGroup


def test_dummy():
    foo = ParameterGroup.from_dict({"foo": [1, 2, 3]})

    foo.get("bar")
```

Running `pytest` will result in this output:
```console
$ pytest dummy_test.py
================================================== test session starts ================================================== 
platform win32 -- Python 3.8.6, pytest-6.2.2, py-1.9.0, pluggy-0.13.1
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibrati

on_precision=10 warmup=False warmup_iterations=100000)
rootdir: D:\git\pyglotaran, configfile: tox.ini
plugins: anyio-2.1.0, nbval-0.9.6, benchmark-3.2.3, cov-2.11.1, env-0.6.2
collected 1 item

[...]

PR #688, opened 2021-06-06
👌 ParameterNotFoundException doesn't preprend '.' if path is empty

s-weigand:
Just a mico change for better error messages I found when writing #687 .

For the error of the dummy example in #687 
Before:
`Cannot find parameter .bar`
After:
`Cannot find parameter 'bar'`

**Testing**


jsnel (approved):
+🩹

PR #697, opened 2021-06-12
🩹 Ensure that the PR number is present in the pr_benchmark_reaction workflow

s-weigand:
Instead of relying on `github.event.workflow_run.pull_requests[0].number` being the proper PR number the workflow ran on, we write it in a file and read it out, the same way we do it with the comment.

### Change summary

- Ensure that the correct PR number is present in the `pr_benchmark_reaction` workflow
- Fixed printing of comment body in the workflow (the problem was the multiline comment string is evaluated before executing the command, which resulted in a syntax error, now it gets wrapped in a python multiline string)

### Checklist

### DIFF
--- a/.github/workflows/pr_benchmark.yml
+++ b/.github/workflows/pr_benchmark.yml
@@ -93,6 +93,7 @@ jobs:
           comment = comment.replace("%", "%25").replace("\n", "%0A").replace("\r", "%0D")
 
           Path("benchmark/.asv/html/pr-diff-comment.txt").write_text(comment)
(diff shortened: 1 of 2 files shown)


jsnel (approved):
Let's hope this works. :)

s-weigand:
> 
> 
> Let's hope this works. :)

Yeah I hope so too, it is about time that it does 😅

s-weigand:
It finally [does work](https://github.com/glotaran/pyglotaran/pull/444#issuecomment-860097056) 🎉

PR #701, opened 2021-06-15
🧹 Remove deprecated call to Model.simulate

s-weigand:
Micro cleanup, of deprecated usage in our code base.

### Change summary

- Change deprecated call to `model.simulate(...)` to `simulate(model, ...)`
- Arguments for `save_result` in the deprecation warning for `Result.save` are in the proper order

### Checklist


jsnel (approved):
👍

PR #704, opened 2021-06-17
🩹 Scale StdError of estimated parameters with RMSE

s-weigand:
In today's meeting with @ism200 we found and solved the problem with the StdError of estimated parameters.

That said following [this StackOverflow aswer](https://stackoverflow.com/a/67023688/3990615) we might want to use a different implementation for cases "when the Jacobian is close to degenerate", but this would need benchmarking first.

Currently we use pretty much use the same as:
```python
chi2dof = np.sum(res.fun**2)/(res.fun.size - res.x.size)
cov *= chi2dof
perr = np.sqrt(np.diag(cov))    # 1sigma uncertainty on fitted parameters
```

Our code (reduced  version):
```python
chi_square = np.sum(ls_result.fun ** 2)
reduced_chi_square = chi_square / degrees_of_freedom
jacobian = ls_result.jac
covariance_matrix = np.linalg.inv(jacobian.T.dot(jacobian))
standard_errors = np.sqrt(np.diagonal(covariance_matrix))
```

The more robust implementation would be:
```python
U, s, Vh = linalg.svd(res.jac, full_matrices=False)
tol = np.finfo(float).eps*s[0]*max(res.jac.shape)
w = s > tol
cov = (Vh[w].T/s[w]**2) @ Vh[w]  # robust covariance matrix
perr = np.sqrt(np.diag(cov))     # 1sigma uncertainty on fitted parameters
```

That might also allow us to get rid of the try-except case [when calculating the `covariance_matrix`](https://github.com/glotaran/pyglotaran/blob/828710e4187bae53709f6bbfe5cc144695d1beed/glotaran/analysis/optimize.py#L112).

### Change summary

- StdError gets scaled with RMSE

### Checklist

- [ ] ✔️ Passing the tests (mandatory for all PR's)
- [ ] 👌 Closes issue (mandatory for ✨ feature and 🩹 bug fix PR's)
- [ ] 📚 Adds documentation of the feature

### Closes issues

closes #703

PR #706, opened 2021-06-18
👌 More robust covariance_matrix calculation

s-weigand:
Follow up PR for #704 , which implements a more robust calculation of the `covariance_matrix` for results.

### Change summary

- 🧪 Adds benchmarks for `glotaran.analysis.optimize._create_result`
- 👌 Uses SVD of the estimated jacobian to calculate `covariance_matrix`

Not sure what was up with the calculation of the mask in the [original StackOverflow answer](https://stackoverflow.com/a/67023688/3990615)

```python
tol = np.finfo(float).eps*s[0]*max(res.jac.shape)
w = s > tol
```

So I changed it to
```python
w = s**2 > np.finfo(float).eps
```
which should be sufficient to prevent `ZeroDivisionError`.

If any of you can think of a reason why the biggest singular value and the shape of the jacobian might be of importance to create a filter mask please tell me 😄 

### Checklist

### DIFF
--- a/benchmark/benchmarks/integration/ex_two_datasets/benchmark.py
+++ b/benchmark/benchmarks/integration/ex_two_datasets/benchmark.py
@@ -1,6 +1,11 @@
+import pickle
 from pathlib import Path
 
+from scipy.optimize import OptimizeResult
+
+from glotaran.analysis.optimize import _create_result
 from glotaran.analysis.optimize import optimize
+from glotaran.analysis.problem_grouped import GroupedProblem
 from glotaran.io import load_dataset
 from glotaran.io import load_model
 from glotaran.io import load_parameters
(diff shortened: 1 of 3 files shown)


s-weigand:
Also if we would refactor `optimize_problem` to return `problem, ls_result, free_parameter_labels, termination_reason` instead of 
`_create_result(problem, ls_result, free_parameter_labels, termination_reason)` and just chain them together in `optimize`, we could get rid of using pickle in the benchmarks which would also make it more maintainable (+ codacy won't complain anymore :P)

jsnel:
We will have to dive a little deeper this time and not only check *if* the examples run, but *how* they run (minimization pathway) and compare the two.

jsnel (approved):
Tested ok!

PR #718, opened 2021-06-25
👌 Changed PR benchmark workflow to run benchmarks from last release

s-weigand:
This will give us a better overview of the changes between versions.

### Change summary

- PR benchmark workflow runs benchmarks from 0.4.0 onward

### Checklist


jsnel (approved):
Good idea. Save some CPU cycles. 🟢 
Also between the first activation of the benchmarks (somewhere around 0.3) and 0.4 there results remained constant anyways.

PR #723, opened 2021-06-26
🚇Fixed starting point of benchamrks to include the 0.4.0 release

s-weigand:
Looks like the starting point of `asv` is exclusive so to benchmark the `0.4.0` release commit as well as the following commits we need to start at `v0.4.0^`.

### Change summary

- `pr_benchmark`workflow now includes the `0.4.0` release commit

### Checklist


jsnel (approved):
Yes.

PR #727, opened 2021-06-29
👌 Fix pre-commit-config

s-weigand:
False usage of the regex might lead to false-positive matching. 'types_or' is the proper way to use pre-commit.

To get `type` of a file you have to use (identify)[https://github.com/pre-commit/identify] (dependency of `pre-commit` so it is already installed)
```console
$ identify-cli <file>
```

### Change summary

- All hooks now use `types_or` to match files

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -124,5 +124,6 @@ repos:
     rev: v2.1.0
     hooks:
       - id: codespell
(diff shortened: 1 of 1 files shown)


jsnel (dismissed):
Ok to merge.

jsnel (approved):
Ok to merge

PR #730, opened 2021-07-02
Feature: Megacomplex Models

joernweissenborn:
This PR removes models like kinetic-image in favor of independent megacomplex model. Therefor, there are a lot of internal changes. 

Key Changes for users:

* use `default-megacomplex` instead of `model-type` in yml spec
* simulations can now be done by setting `global_megacomplex` in a dataset model
* heterogenous dataset models can be analyzed in one optimization, e.g. a spectral and a temporal dataset
* the user now needs to set `group=true` in a scheme to have a grouped analysis.

### Checklist

- [x] ✔️ Passing the tests (mandatory for all PR's)
- [ ] 👌 Closes issue (mandatory for ✨ feature and 🩹 bug fix PR's)
- [x] 🧪 Adds new tests for the feature (mandatory for ✨ feature and 🩹 bug fix PR's)
- [x] 📚 Adds documentation of the feature
- [x] Added Deprecations

### Closes issues

none

PR #733, opened 2021-07-03
👌 Monkeypatch 'scheme.group = True' in benchmark

s-weigand:
With the changed default behavior in PR730 , forcing `scheme.group = True` allows the benchmark to run consistently.

### Change summary

- Benchmarks will consistently do the same optimization after #730 got merged

### Checklist


joernweissenborn (commented):
lgtm

jsnel:
We decided (sign-off: @s-weigand) this is not needed, we will make scheme.group = True the default in case of multiple datasets, since this was also the (sensible) default in TIMP. This has the added advantage that the examples (integration tests) do not need to change and no monkeypatch commit is needed on main.

PR #740, opened 2021-07-08
🩹 Fix Performance Regressions

joernweissenborn:
This PR Addresses performance regressions introduce in staging branch.

* Adds pytest benchmark for `analysis.Problem` class
*Adds numerous performance tweaks
* Regressions: some internal objects are no longer available via public API, namely the reduced clps.

### Checklist

### Closes issues

None


s-weigand:
I think those 3 functions could be one since [`benchmark` returns the result](https://github.com/ionelmc/pytest-benchmark/#examples).
`result = benchmark(something)`

glotaran/analysis/util.py
```diff
@@ -53,66 +62,78 @@ def calculate_matrix(
...
     if global_model:
         dataset_model.swap_dimensions()
 
-    return matrix
+    return CalculatedMatrix(clp_labels, matrix)
+
+
+@nb.jit(nopython=True, parallel=True)
+def apply_weight(matrix, weight):
+    for i in range(matrix.shape[1]):
```
s-weigand, line 92:
```suggestion
    for i in nb.prange(matrix.shape[1]):
```
Not sure if `numba` does that already because of the `parallel=True`, but it is better to make sure of it.

jsnel:
A note on the spectral model validation

Results for [paramGUI case 05](https://github.com/glotaran/pyglotaran-validation/tree/main/paramGUI/simData05)

Compare the estimated parameters from pyglotaran 
```
  * __spectral__:
    | _Label_   |      _Value_ |    _StdErr_ |
    |-----------|--------------|-------------|
    | amp1      |     1        | 0           |
    | loc1      | 14282        | 0.0193176   |
    | width1    |   808.004    | 0.0343336   |
    | skew1     |     0.400081 | 8.92553e-05 |
    | amp2      |     1        | 0           |
    | loc2      | 13707        | 0.0331115   |
    | width2    |   654.987    | 0.0556831   |
    | skew2     |    -0.299748 | 0.000158642 |
```
with those from paramGUI as recored in specFitSummary.txt:
Parameters:
```
     Estimate Std. Error t value Pr(>|t|)    
t1  1.428e+04  1.971e-02  724710   <2e-16 ***
t2  8.080e+02  3.503e-02   23069   <2e-16 ***
t3  4.001e-01  9.106e-05    4394   <2e-16 ***
t4  1.371e+04  3.378e-02  405740   <2e-16 ***
t5  6.550e+02  5.681e-02   11529   <2e-16 ***
t6 -2.997e-01  1.619e-04   -1852   <2e-16 ***
```
Also note the similarity in plotted results: [paramGUI](https://github.com/glotaran/pyglotaran-validation/blob/main/paramGUI/simData05/specFit.png) vs [pyglotaran](https://github.com/glotaran/pyglotaran-validation/blob/main/paramGUI/simData05/spectral_model_fit_pyglotaran.png)
So you see the values estimated are essentially the same (when units are compatible).

By compatible I mean that the units of the spectral model parameters are the same as the spectral axis of the dataset (in this case both `cm^-1`), this is not possible in all cases (even with the option energy_spectrum: True). 
Some mechanism is needed which not only allows one to specify whether a megacomplex is 'energy-spectrum' but also if the dataset is 'energy-spectrum', so you can 'cancel out' if they are already compatible and only convert when needed. In the future some kind of `type-annotation` of units is needed (`nm`, `cm^-1`, `eV`, etc) for all data-types: dataset axis, model axis, parameters.

For now the results are good enough to merge into staging, at least from this perspective.

glotaran/analysis/util.py
```diff
@@ -160,55 +179,78 @@ def retrieve_clps(
...
+        if interval[0] > global_axis[-1]:
+            continue
+
+        start_idx, end_idx = get_idx_from_interval(interval, global_axis)
+        for i in range(start_idx, end_idx + 1):
+            index_clp_labels = clp_labels[i] if isinstance(clp_labels[0], list) else clp_labels
+            if clp_label in index_clp_labels:
+                area.append(clps[i][index_clp_labels.index(clp_label)])
+
+    return np.asarray(area)  # TODO: normalize for distance on global axis
```
s-weigand, line 241:
```suggestion
    return np.sum(area)  # TODO: normalize for distance on global axis
```
`area` being a float makes more sense and we would need to call `np.sum` on it in `calculate_clp_penalties`.
Same is true for `_get_area` in `glotaran/model/clp_penalties.py`.

jsnel (approved):
Taking note of [my comment](https://github.com/glotaran/pyglotaran/pull/740#issuecomment-879490420) w.r.t the spectral model validation, this PR is good enough to merge to staging,
pending possibly some more detailed general comments by @s-weigand.

s-weigand (changes requested):
Additional issues I found.

https://github.com/glotaran/pyglotaran/blob/cf16572ae93d2fec1979bcb348fb8c05e1758c8d/glotaran/analysis/util.py#L44

`indices` should have type `dict[str, int]` not  `dict[str, int] | None`, since the mega completes in `builtin` have that signature and don't defend against `None`.

joernweissenborn:
> Additional issues I found.
> 
> https://github.com/glotaran/pyglotaran/blob/cf16572ae93d2fec1979bcb348fb8c05e1758c8d/glotaran/analysis/util.py#L44
> 
> `indices` should have type `dict[str, int]` not  `dict[str, int] | None`, since the mega complexes in `builtin` have that signature and don't defend against `None`.

Right, i will fix that.

s-weigand:
Last benchmark before review suggestions got applies
```
All benchmarks:

       before           after         ratio
     [dc00e6da]       [8cc6e6e0]
     <v0.4.0>                   
-      58.2±0.5ms       39.7±0.9ms     0.68  BenchmarkOptimize.time_optimize(False, False, False)
-         275±2ms         128±40ms     0.46  BenchmarkOptimize.time_optimize(False, False, True)
         82.9±3ms         67.2±2ms    ~0.81  BenchmarkOptimize.time_optimize(False, True, False)
+      82.4±0.9ms         128±20ms     1.55  BenchmarkOptimize.time_optimize(False, True, True)
-      58.6±0.9ms         48.5±1ms     0.83  BenchmarkOptimize.time_optimize(True, False, False)
-         267±5ms         109±30ms     0.41  BenchmarkOptimize.time_optimize(True, False, True)
         83.1±1ms         78.3±2ms     0.94  BenchmarkOptimize.time_optimize(True, True, False)
+        84.2±2ms         144±10ms     1.71  BenchmarkOptimize.time_optimize(True, True, True)
             183M             180M     0.98  IntegrationTwoDatasets.peakmem_create_result
             202M             198M     0.98  IntegrationTwoDatasets.peakmem_optimize
-         255±3ms          200±4ms     0.78  IntegrationTwoDatasets.time_create_result
-      5.43±0.02s       1.50±0.08s     0.28  IntegrationTwoDatasets.time_optimize
```

glotaran/analysis/util.py
joernweissenborn: Nope, it is really needed for the the optimization. Each penalty is not just added to the sum, but rather increases the size of the penalty vector.

joernweissenborn: Ok, after personal discussion with @s-weigand we decided wo make a decicated issue for that.

s-weigand: https://github.com/glotaran/pyglotaran/issues/746

PR #745, opened 2021-07-10
🔧 Mock '_create_result' in benchmarks

s-weigand:
This PR just demonstrates the influence of `glotaran.analysis.optimize._create_result` on the benchmarks, else it is the same as #740 

To reduce the noise in `optimize` caused by `glotaran.analysis.optimize._create_result` we could simply mock it.


jsnel:
No longer needed.

PR #747, opened 2021-07-15
Feature: Full Models

### LINKED ISSUE #750: ✨ Implement spectrotemporal models as implemented in paramGUI
A spectrotemporal model for fitting time-resolved spectroscopy data is described in the paper
**Global and target analysis of time-resolved spectra** DOI: [10.1016/j.bbabio.2004.04.011](https://doi.org/10.1016/j.bbabio.2004.04.011)
 
A public implementation is available in the R package [paramGUI](https://github.com/glotaran/paramGUI).

This issue is about implementing equivalent functionality, but implemented using megacomplex composition.

A repository for validation example code can be found here: [pyglotaran-validation](https://github.com/glotaran/pyglotaran-validation/) this contains the (reference) [paramGUI examples](https://github.com/glotaran/pyglotaran-validation/tree/main/paramGUI) to test the implantation against.

joernweissenborn:
This PR adds the ability to optimize full models to glotaran. 

### Change summary

- Full models can be fitted
- ~~Only non index dependent models are supported for now (might change till finishof this PR)~~
- No relations and constraints are supported on either the global and model matrix
- Clp penalties are not supported

### Checklist

- [x] fix my reviews
- [x] put guards to restrain the full model (no constraints eg)
- [x] implement index dependent model matrix for full model

### Closes

closes #750

### DIFF
--- a/glotaran/analysis/problem_ungrouped.py
+++ b/glotaran/analysis/problem_ungrouped.py
@@ -172,8 +172,8 @@ class UngroupedProblem(Problem):
             self._additional_penalty.append(additional_penalty)
 
     def _calculate_full_model_residual(self, label: str, dataset_model: DatasetModel):
-        model_matrix = self.matrices["dataset1"].matrix
-        global_matrix = self.global_matrices["dataset1"].matrix
+        model_matrix = self.matrices[label].matrix
+        global_matrix = self.global_matrices[label].matrix
 
         matrix = np.kron(global_matrix, model_matrix)
         weight = self._flattened_weights.get(label)
--- a/glotaran/analysis/problem_ungrouped.py
+++ b/glotaran/analysis/problem_ungrouped.py
@@ -12,10 +12,36 @@ from glotaran.analysis.util import calculate_matrix
 from glotaran.analysis.util import reduce_matrix
 from glotaran.analysis.util import retrieve_clps
 from glotaran.model import DatasetModel
+from glotaran.project import Scheme
 
 
 class UngroupedProblem(Problem):
     """Represents a problem where the data is not grouped."""
+
+    def __init__(self, scheme: Scheme):
+        """Initializes the Problem class from a scheme (:class:`glotaran.analysis.scheme.Scheme`)
+
+        Args:
+            scheme (Scheme): An instance of :class:`glotaran.analysis.scheme.Scheme`
+                which defines your model, parameters, and data
+        """
+        super().__init__(scheme=scheme)
+
+        self._global_matrices = {}
+        self._flattened_data = {}
+        self._flattened_weights = {}
+        for label, dataset_model in self.dataset_models.items():
+            if dataset_model.global_model():
+                self._flattened_data[label] = dataset_model.get_data().T.flatten()
+                weight = dataset_model.get_weight()
+                if weight is not None:
+                    weight = weight.T.flatten()
+                    self._flattened_data[label] *= weight
+                    self._flattened_weight[label] = weight
+
+    @property
+    def global_matrices(self) -> dict[str, CalculatedMatrix]:
+        return self._global_matrices
 
     def calculate_matrices(
         self,
@@ -28,6 +54,7 @@ class UngroupedProblem(Problem):
             raise ParameterError
 
         self._matrices = {}
+        self._global_matrices = {}
         self._reduced_matrices = {}
 
         for label, dataset_model in self.dataset_models.items():
@@ -36,6 +63,9 @@ class UngroupedProblem(Problem):
                 self._calculate_index_dependent_matrix(label, dataset_model)
             else:
                 self._calculate_index_independent_matrix(label, dataset_model)
+
+            if dataset_model.global_model:
+                self._calculate_global_matrix(label, dataset_model)
 
         return self._matrices, self._reduced_matrices
 
@@ -52,11 +82,14 @@ class UngroupedProblem(Problem):
             self._reduced_matrices[label].append(reduced_matrix)
 
     def _calculate_index_independent_matrix(self, label: str, dataset_model: DatasetModel):
-
         matrix = calculate_matrix(dataset_model, {})
         self._matrices[label] = matrix
         reduced_matrix = reduce_matrix(matrix, self.model, self.parameters, None)
         self._reduced_matrices[label] = reduced_matrix
+
+    def _calculate_global_matrix(self, label: str, dataset_model: DatasetModel):
+        matrix = calculate_matrix(dataset_model, {}, global_model=True)
+        self._global_matrices[label] = matrix
 
     def calculate_residual(
         self,
@@ -75,7 +108,10 @@ class UngroupedProblem(Problem):
         self._additional_penalty = []
 
         for label, dataset_model in self._dataset_models.items():
-            self._calculate_residual(label, dataset_model)
+            if dataset_model.global_model():
+                self._calculate_full_model_residual(label, dataset_model)
+            else:
+                self._calculate_residual(label, dataset_model)
 
         self._additional_penalty = (
             np.concatenate(self._additional_penalty) if len(self._additional_penalty) != 0 else []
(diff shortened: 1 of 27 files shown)


glotaran/analysis/problem_ungrouped.py
```diff
@@ -37,6 +64,9 @@ def calculate_matrices(
             else:
                 self._calculate_index_independent_matrix(label, dataset_model)
 
+            if dataset_model.global_model:
```
joernweissenborn, line 67: this is a bug.

joernweissenborn: `global_model` is a function

jsnel:
Thank you for these nice ✨ 

--

Please find the python scripts uploaded to the [pyglotaran-validation](https://github.com/glotaran/pyglotaran-validation/) repository.

In particular [this commit](https://github.com/glotaran/pyglotaran-validation/commit/6ffd12c869dce54d37fb0b1d5a61bd780b8ef089) contains the update to test the full model functionality on two basic examples. 

See the target results in:

- [paramGUI/simData01/specTempFitSummary.txt](https://github.com/glotaran/pyglotaran-validation/blob/main/paramGUI/simData01/specTempFitSummary.txt)
- [paramGUI/simData05/specTempFitSummary.txt](https://github.com/glotaran/pyglotaran-validation/blob/main/paramGUI/simData05/specTempFitSummary.txt)

And congratulations on this first step in reaching the milestone of bringing Spectrotemporal ('Full') models to pyglotaran! 👏

glotaran/builtin/megacomplexes/spectral/spectral_megacomplex.py
```diff
@@ -61,18 +68,23 @@ def finalize_data(self, dataset_model: DatasetModel, data: xr.Dataset):
                     compartment for compartment in megacomplex.shape if compartment not in species
                 ]
 
-        data.coords["species"] = species
+        data.coords[species_dimension] = species
+        matrix = data.global_matrix if as_global else data.matrix
```
jsnel, line 72: This lines currently crashes example [fit_simData05.py](https://github.com/glotaran/pyglotaran-validation/blob/main/paramGUI/simData05/fit_simData05.py)

glotaran/test/test_spectral_decay_full_model.py
```diff
@@ -0,0 +1,223 @@
...
+    mc1:
+        type: decay
+        k_matrix: [k1]
+    mc2:
+        type: spectral
+        shape:
+            s1: sh1
+            s2: sh2
+            s3: sh3
+irf:
```
jsnel, line 38: Can you add a case without IRF? Currently the full_model doesn't work without (see [fit_simData01.py](https://github.com/glotaran/pyglotaran-validation/blob/main/paramGUI/simData01/fit_simData01.py))

jsnel (changes requested):
Please add a test case with no IRF and review the finalize_data method in spectral_megacomplex.py, as indicated in the comments.

glotaran/analysis/problem_ungrouped.py
s-weigand: Yeah, this is due to the nature of how `if` it checks for truthiness.
From the Zen of python (`import this`):
> Explicit is better than implicit.

So using `==` consistently can prevent this bug.
```python
In [1]: def func():
      :     return True
      :

In [2]: if func:
      :     print(True)
      : else:
      :     print(False)
      :
True

In [3]: if func():
      :     print(True)
      : else:
      :     print(False)
      :
True

In [4]: if func == True:
      :     print(True)
      : else:
      :     print(False)
      :
False

In [5]: if func() == True:
      :     print(True)
      : else:
      :     print(False)
      :
True
```

Something else to be aware of is that for historical reasons `bool` is a subclass of `int`:
```python
In [6]: if func() == 1:
      :     print(True)
      : else:
      :     print(False)
      :
True

In [7]: if True == 1:
      :     print(True)
      : else:
      :     print(False)
      :
True
```

So if a function/method can return `bool` or `int` and you need to treat them differently you should use.
```python
In [8]: if isinstance(func(), bool) and func() == True:
       :     print(True)
       : else:
       :     print(False)
       :
True
```

glotaran/test/test_spectral_decay_full_model.py
```diff
@@ -0,0 +1,223 @@
...
+class ThreeComponentSequential:
+    model = load_model(MODEL_3C_SEQUENTIAL, format_name="yml_str")
+    initial_parameters = load_parameters(PARAMETERS_3C_INITIAL_SEQUENTIAL, format_name="yml_str")
+    wanted_parameters = load_parameters(PARAMETERS_3C_SIM_SEQUENTIAL, format_name="yml_str")
+    time = np.arange(-10, 50, 1.0)
+    spectral = np.arange(600, 750, 5.0)
+    axis = {"time": time, "spectral": spectral}
+
+
+@pytest.mark.skip("Not working yet")
```
jsnel, line 166: TODO: making it work

glotaran/analysis/problem_ungrouped.py
jsnel: I think this issue has been resolved, in all instances it is called correctly now:
image
So I'll mark this as resolved.

joernweissenborn (commented):
On second thought this is needed so it so the megacomplex knows when to look for `global_matrix`

glotaran/builtin/megacomplexes/spectral/spectral_megacomplex.py
joernweissenborn: Thats fixed

glotaran/test/test_spectral_decay_full_model.py
joernweissenborn: So the problem is that the full models seem to be more sensible in paramters. There was a bug which led to that the result data was calculated with the parameters of last function eval, not the optimized. Ifixed this and it should work now. See also the unittest I added.

glotaran/test/test_spectral_decay_full_model.py
joernweissenborn: Works now

glotaran/project/scheme.py
```diff
@@ -75,6 +76,15 @@ def markdown(self):
 
         return MarkdownStr(markdown_str)
 
+    def grouped(self) -> bool:
```
s-weigand, line 79: Nitpicky.
Renaming it to `is_goupable` would make more sense.
Could be done in a different PR.

glotaran/test/test_spectral_decay.py
```diff
@@ -7,6 +7,50 @@
 from glotaran.io import load_parameters
 from glotaran.project import Scheme
 
+MODEL_3C_NO_IRF = """\
+dataset:
+    dataset1: &dataset1
```
s-weigand, line 12:
```suggestion
    dataset1:
```
The anchor isn't used, so there is no need to define it.

glotaran/test/test_spectral_decay_full_model.py
```diff
@@ -0,0 +1,224 @@
...
+    dataset2: &dataset2
+        megacomplex: [mc2]
+        global_megacomplex: [mc1]
+        initial_concentration: j1
+        irf: irf1
+    dataset3: &dataset3
+        megacomplex: [mc1]
+        initial_concentration: j1
+        irf: irf1
+    dataset4: &dataset4
```
s-weigand, line 26:
```suggestion
    dataset1:
        megacomplex: [mc1]
        global_megacomplex: [mc2]
        initial_concentration: j1
        irf: irf1
    dataset2:
        megacomplex: [mc2]
        global_megacomplex: [mc1]
        initial_concentration: j1
        irf: irf1
    dataset3:
        megacomplex: [mc1]
        initial_concentration: j1
        irf: irf1
    dataset4:
```
The anchors aren't used, so there is no need to define them.

s-weigand (changes requested):
Mostly nitpicky things that caught my eye and can be done in a different PR (given that @jsnel is also fine with the current changes).
I also opened an issue for the boolean switch naming things #755

jsnel (approved):
Full model validated against simulation in pyglotaran-simulations (cross checked with TIMP/paramGUI).
Existing functionality checked against all examples in pyglotaran-examples.

Some loss of typing information found, to be addressed in a future PR.

s-weigand (approved):
I trust @jsnel on this ^^

PR #754, opened 2021-07-17
🧪🚇 Add integration test result validation

### LINKED ISSUE #753: 🚇Add result validation step to CI
### Describe the tooling

Right now the integration tests only show breaking API changes, but checking validity is still a manual task.
The CI should show us if the results change and we should investigate further.

s-weigand:
Right now the integration tests only show breaking API changes, but checking validity is still a manual task.
This PR adds a way to compare the results to a "gold standard" (tag, branch, commit) set by [this workflow](https://github.com/glotaran/pyglotaran-examples/actions/workflows/update_compare_results.yml) on the `pyglotaran-examples` repo.

### Change summary

- Adds numerical comparison to "gold standard" results defined by `pyglotaran-examples`

### Checklist

### Closes issues

closes #753

### DIFF
--- a/.github/test_result_consistency.py
+++ b/.github/test_result_consistency.py
@@ -1,0 +1,148 @@
+""""Tests to ensure result consistency."""
+from __future__ import annotations
+
+import os
+import subprocess
+from collections import defaultdict
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING
+from warnings import warn
+
+import numpy as np
+import pytest
+import xarray as xr
+
+if TYPE_CHECKING:
+    from xarray.core.coordinates import DataArrayCoordinates
+
+
+REPO_ROOT = Path(__file__).parent.parent
+RUN_EXAMPLES_MSG = (
+    "run 'python scripts/run_examples.py run-all' in the 'pyglotaran-examples' repo root."
+)
+
+
+def get_compare_results_path() -> Path:
+    """Ensure that the comparison-results exist, are up to date and return their path."""
+    compare_result_folder = REPO_ROOT / "comparison-results"
+    if not compare_result_folder.exists():
+        example_repo = "git@github.com:glotaran/pyglotaran-examples.git"
+        subprocess.run(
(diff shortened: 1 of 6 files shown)


s-weigand:
For this workflow to pass we might need to fix how `problem.save_parameters_for_history` is implemented in `optimize` and make a fix release `0.4.1`, since it is done wrong according to @joernweissenborn 🤷‍♀️.
At least we now have a way to compare the results across versions.

jsnel (approved):
LGTM

s-weigand:
The `clp_label` swapping for `ex_two_datasets` in [this CI run](https://github.com/glotaran/pyglotaran/runs/3329625935?check_suite_focus=true)
might be related to the implementation of `involved_compartments` 
https://github.com/glotaran/pyglotaran/blob/d2b297b534dd7092eb3c00d69132da7c0e4971ee/glotaran/builtin/megacomplexes/decay/k_matrix.py#L47

Since set **does not** preserve order.

Using `frozenset` might solve this, or even better using the user-provided order from the defined model (not sure if we still have access to that on this point).

s-weigand:
@jsnel  and I did a long debugging session today where we found the following:
- weights are applied differently (compartment order) between `0.4.1` and `staging` (failing `ex_two_datasets` and possibly `simultaneous_analysis_6d_disp`)
- the missing coordinate `spectral` is due to different code paths when auto determining index dependency `UngroupedProblem.create_index_dependent_result_dataset` vs. `UngroupedProblem.create_index_independent_result_dataset` (also this is kinda random or maybe I messed the tests up [Missing coordinate: 'spectral' in 'dataset1.nc', data_var 'matrix'](https://github.com/glotaran/pyglotaran/runs/3334331574?check_suite_focus=true) vs. [`spectral` is apparently present](https://github.com/glotaran/pyglotaran/pull/754/checks?check_run_id=3335789476))
- problems in `simultaneous_analysis_6d_disp` might be due to missing noise and/or the differences in how weights are applied

PR #760, opened 2021-07-31
🧪🚇 Add integration test result validation (maintain/0.4.1)

s-weigand:
Right now the integration tests only show breaking API changes, but checking validity is still a manual task.
This PR adds a way to compare the results to a "gold standard" (tag, branch, commit) set by [this workflow](https://github.com/glotaran/pyglotaran-examples/actions/workflows/update_compare_results.yml) on the `pyglotaran-examples` repo.

### Change summary

- Adds numerical comparison to "gold standard" results defined by `pyglotaran-examples`

### Checklist

### DIFF
--- a/.github/test_result_consistency.py
+++ b/.github/test_result_consistency.py
@@ -1,0 +1,148 @@
+""""Tests to ensure result consistency."""
+from __future__ import annotations
+
+import os
+import subprocess
+from collections import defaultdict
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING
+from warnings import warn
+
+import numpy as np
+import pytest
+import xarray as xr
+
+if TYPE_CHECKING:
+    from xarray.core.coordinates import DataArrayCoordinates
+
+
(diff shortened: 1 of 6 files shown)


s-weigand:
We (@jsnel and @s-weigand ) figured out which examples need to be reviewed (`EXAMPLE_BLOCKLIST`) and that the SVD (which is just used for tooling) gives different results on windows vs. linux and thus needs a less strict threshhold.

jsnel (approved):
LGTM. I think we have tested this enough.

There's still the residual matter of the results not being exactly the same depending on the environment it's run in (and sometimes even consecutive runs of the CI) but this is a different matter that can always be resolved on a per case basis by manual inspection if needed. The tolerances set now are now at least enough to not trigger to many false positives on the CI and at least catch true positives (as in some regression).

jsnel:
Keeping this PR open as a sanity test for a similar PR to staging for the "Compare results" feature.

jsnel (approved):
👌

PR #762, opened 2021-08-07
Small fix for baseline megacomplex

joernweissenborn:
Really small fix


s-weigand (approved):
LGTM

jsnel (approved):
Looks ok. We currently don't have models with baseline as part of examples so this change should not break anything.

PR #763, opened 2021-08-08
Refactor/spectral model

joernweissenborn:
This PR introduces some changes to the spectral model in order to "straighten it out" and enable TIMP like behaviour.

### Change summary

- Introduced `SpectralShapeGaussian` as shape type `gaussian`
- Made `skewness` property mandatory for shape `skewed-gaussian`
- removed `energy_spectrum` property from spectral megacomplex
- added `spectral_axis_inverted` and `spectral_axis_scale` to dataset model (see below)

#### Changes to spectral axis

The old behaviour was to invert the axis as `axis = 1e7 / axis` if `energy_spectrum` was set true in the corresponding megacomplex.

The new behaviour inverts the axis as `axis = dataset_model.spectral_axis_scale / axis` if `spectral_axis_inverted` property of the dataset model is set true.  The `spectral_axis_scale` is also a dataset model property and is default `1`.  If the inverted property is not set and scale != `1` is provided, the axis will be scaled as `axis = dataset_model.spectral_axis_scale * axis`

### Checklist

### DIFF
--- a/glotaran/builtin/megacomplexes/spectral/shape.py
+++ b/glotaran/builtin/megacomplexes/spectral/shape.py
@@ -9,7 +9,7 @@ from glotaran.parameter import Parameter
 
 @model_item(
     properties={
-        "amplitude": Parameter,
+        "amplitude": {"type": Parameter, "allow_none": True},
         "location": Parameter,
         "width": Parameter,
     },
@@ -55,9 +55,10 @@ class SpectralShapeGaussian:
         np.ndarray
             An array representing a Gaussian shape.
         """
-        return self.amplitude * np.exp(
-            -np.log(2) * np.square(2 * (axis - self.location) / self.width)
-        )
+        shape = np.exp(-np.log(2) * np.square(2 * (axis - self.location) / self.width))
+        if self.amplitude is not None:
+            shape *= self.amplitude
+        return shape
 
 
 @model_item(
@@ -124,12 +125,14 @@ class SpectralShapeSkewedGaussian(SpectralShapeGaussian):
         if np.allclose(self.skewness, 0):
             return super().calculate(axis)
         log_args = 1 + (2 * self.skewness * (axis - self.location) / self.width)
-        result = np.zeros(log_args.shape)
+        shape = np.zeros(log_args.shape)
         valid_arg_mask = np.where(log_args > 0)
-        result[valid_arg_mask] = self.amplitude * np.exp(
+        shape[valid_arg_mask] = np.exp(
             -np.log(2) * np.square(np.log(log_args[valid_arg_mask]) / self.skewness)
         )
-        return result
+        if self.amplitude is not None:
+            shape *= self.amplitude
+        return shape
 
 
 @model_item(properties={}, has_type=True)
--- a/glotaran/builtin/megacomplexes/spectral/shape.py
+++ b/glotaran/builtin/megacomplexes/spectral/shape.py
@@ -124,6 +124,11 @@ class SpectralShapeSkewedGaussian:
         np.ndarray
             An array representing a skewed Gaussian shape.
         """
+        if np.allclose(self.skewness, 0):
+            raise ValueError(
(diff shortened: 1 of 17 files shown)


glotaran/builtin/megacomplexes/spectral/shape.py
```diff
@@ -12,49 +12,13 @@
...
-        Internally ``axis`` is converted from :math:`\mbox{nm}` to
-        :math:`1/\mbox{cm}`, thus ``location`` and ``width`` also need to
-        be provided in :math:`1/\mbox{cm}` (``1e7/value_in_nm``).
-
-        """
-        return (
-            self.calculate_skewed_gaussian(axis)
-            if self.skewness is not None and not np.allclose(self.skewness, 0)
-            else self.calculate_gaussian(axis)
-        )
```
s-weigand, line 55: This part is needed in the calculation of `SpectralShapeSkewedGaussian`, to prevent `ZeroDevisionError` in cases where `self.skewness` is close to zero.
If `SpectralShapeSkewedGaussian` would inherit from `SpectralShapeGaussian` you could use `super().calculate(axis)`.

s-weigand: We should add a test with a small skewness where the initial parameter has the wrong sign and thus changes it during fitting.
Or at least a unit test for `SpectralShapeSkewedGaussian` with `self.skewness = 0` where the results are the same as for `SpectralShapeGaussian`.

joernweissenborn: As mentioned numerous times: This is not a bug, it is a feature. In that case the user must switch to a gaussian model in that cases. The skew is a more advanced feature and people need to be aware of it. I also had this in discussion with @ism200 who agrees that we should not protect against this. But I will put in an Exception for this case.

joernweissenborn: I added the exception 98b6b66

joernweissenborn: I reverted to old behavior.

glotaran/builtin/megacomplexes/spectral/shape.py
```diff
@@ -95,7 +59,17 @@ def calculate_gaussian(self, axis: np.ndarray) -> np.ndarray:
...
+    properties={
+        "skewness": Parameter,
+    },
+    has_type=True,
+)
+class SpectralShapeSkewedGaussian(SpectralShapeGaussian):
+    """A skewed Gaussian spectral shape"""
+
+    def calculate(self, axis: np.ndarray) -> np.ndarray:
         r"""Calculate the skewed Gaussian shape for ``axis``.
```
s-weigand, line 73: The docstring needs to be updated in line 111 to point at `SpectralShapeGaussian.calculate` instead of the old `calculate_gaussian`.

s-weigand (changes requested):
Besides the requested change to the docstring, this LGTM.

s-weigand:
Also, one more minor side note, don't use blank `# noqa`, use `# noqa: T001` instead.
Or even better add a per file ignore to the flake8 config (`tox.ini` L22), that way you don't need to use it over and over again.
```ini
    # Allow printing in test files
    test_*.py: T001
```

glotaran/builtin/megacomplexes/spectral/shape.py
joernweissenborn: Fixed.

joernweissenborn:
> Also, one more minor side note, don't use blank `# noqa`, use `# noqa: T001` instead.
> Or even better add a per file ignore to the flake8 config (`tox.ini` L22), that way you don't need to use it over and over again.
> 
> ```ini
>     # Allow printing in test files
>     test_*.py: T001
> ```

We should do this in a followup PR where we add flake8-print to pre-commit.

s-weigand (approved):
Should be fine now

jsnel (approved):
Looks good to me now.

PR #785, opened 2021-08-22
🚇 Speedup PR benchmark

s-weigand:
Running benchmarks on PRs takes forever to get feedback (~1h e.g. [here](https://github.com/glotaran/pyglotaran/runs/3382108255)) since it benchmarks each commit in between.
The most important information are the benchmark of the last release and the current commit before merging.
Thus we can omit running the benchmarks for all steps in between.
If more insight is needed at which specific commit in a PR a performance regression occurred it can always be run locally.

### Change summary

- Only run base compare commit (`v0.4.0`) and last `HEAD` commit on PR benchmark

### Checklist

### DIFF
--- a/.github/workflows/pr_benchmark.yml
+++ b/.github/workflows/pr_benchmark.yml
@@ -48,7 +48,8 @@ jobs:
           git remote add upstream https://github.com/glotaran/pyglotaran
           git fetch upstream
           pushd benchmark
-          asv run v0.4.0^..HEAD --machine gh_action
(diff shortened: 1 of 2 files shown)


s-weigand:
The benchmarking time is down to [~3min](https://github.com/glotaran/pyglotaran/pull/785/checks?check_run_id=3392671360) 🎉

jsnel (approved):
👍 🕐++

PR #790, opened 2021-08-23
Update calls to clp_area_penalties where deprecated func is called

### LINKED ISSUE #789: 🐛 Fix calling of clp_area_penalties instead of deprecated equal_area_penalties
### Version information

- pyglotaran version: staging

### To Reproduce

1. Run '...'

```python
from glotaran import ...
```

1. See error

jsnel:
The deprecated equal_area_penalties would be called in lieu of the new clp_area_penalties function

The pyglotaran-examples have in their model still equal_area_penalties which is correctly swapped with clp_area_penalties but them model.equal_area_penalties  was still called in some places instead of model.clp_area_penalties

### Checklist

### Closes issues

closes #789

### DIFF
--- a/glotaran/model/clp_penalties.py
+++ b/glotaran/model/clp_penalties.py
@@ -60,7 +60,7 @@ class EqualAreaPenalty:
 
 
 def has_spectral_penalties(model: Model) -> bool:
-    return len(model.equal_area_penalties) != 0
+    return len(model.clp_area_penalties) != 0
 
 
 def apply_spectral_penalties(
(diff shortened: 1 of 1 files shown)


jsnel:
Ignore the codecov result [comment](https://github.com/glotaran/pyglotaran/pull/790#issuecomment-904184261) which fails because the commit change single line function which are not under coverage so the relative change is through the roof even though the code coverage in absolute sense doesn't change.

joernweissenborn (approved):
Typical joern copy paste bugs, thx for fixing

PR #775, opened 2021-08-13
🩹Fix check_deprecations not showing deprecation warnings

s-weigand:
To have a smooth transition for users the changes in the model specs were facilitated by `check_deprecations`
https://github.com/glotaran/pyglotaran/blob/bef7262348b9385b5abd1b2edf260b35ebaccc8a/glotaran/utils/sanitize.py#L221
which replaces the deprecated dict keys and values of the old spec with the new ones internally and should warn the user about the changed usage.
But the deprecation warnings were [filtered by python since the stack level didn't reach "usercode" (`load_model`)](https://docs.python.org/3/library/warnings.html#default-warning-filter).

This PR fixes this and adds tests that the warning originates from `load_model`, so users will see it.

Also, instead of `DeprecationWarning` we now raise `GlotaranApiDeprecationWarning` which is a subclass of `UserWarning` so it won't be filtered by python.

### Change summary

- ✨ Added 'deprecate_dict_entry' to deprecate dict keys and/or values
- 🧹📚 Changed Warns OverDueDeprecation to Raises and added missing 
- 🩹🧪 Reimplemented check_deprecations and with deprecate_dict_entry 
- ♻️🧪 Changed deprecation_warning_on_call_test_helper to test for file 
- 📚 Added docs for deprecate_dict_entry ([link](https://pyglotaran--775.org.readthedocs.build/en/775/contributing.html#deprecating-dict-entries))
- 👌 Changed DeprecationWarning to GlotaranApiDeprecationWarning

### Checklist

- [ ] ✔️ Passing the tests (mandatory for all PR's)
- [ ] 👌 Closes issue (mandatory for ✨ feature and 🩹 bug fix PR's)
- [x] 🧪 Adds new tests for the feature (mandatory for ✨ feature and 🩹 bug fix PR's)
- [x] 📚 Adds documentation of the feature

PR #795, opened 2021-08-29
🩹 Fix wrong deprecation replacement for spectral-model spec

s-weigand:
Follow-up PR for #775 .
The replacement wor backward compatibility (`type: spectrum` -> `default-megacomplex: spectral`) in #775 was wrong, since the old way (`v0.4.0`) to define a spectral model was `type: spectral-model`.
This error slipped since we do not yet have integration tests for spectral models/megacomplexes.

### Change summary

- Fixes broken backward compatibility for `0.4.0` spectral model specs.

### Checklist


jsnel (approved):
LGTM

PR #801, opened 2021-08-29
Improvements to application of clp_penalties (equal area)

jsnel:
This fix reproduces the area penalty behavior that was in v0.4.1 when multiple datasets are involved.

The 'equal area penalty' is calculated on a per dataset basis.

### Change summary

- Calculate in `calculate_clp_penalties` the clp penalties on a per dataset level

### Checklist

- [x] ✔️ Passing the tests (mandatory for all PR's)
- [x] 👌 Closes issue (mandatory for ✨ feature and 🩹 bug fix PR's)

### Closes issues

closes #800

Issue #800, opened 2021-08-29
🐛🤔♻️ Decide how to apply spectral penalties with multiple datasets and/or different spectral ranges


jsnel:
### Expected behavior

TODO

### Screenshots

TODO

### Additional context

TODO

jsnel:
A decision on this needs to be made with input from @ism200, to be discussed in a future project meeting.

s-weigand:
See also #746

PR #802, opened 2021-09-01
Experiments/compare results with penalty scaling

jsnel:
Experimental PR combining #754 with #801

### Change summary

- See #754 
- See #801 

### Checklist


jsnel:
Experiment was successful, this PR showed that #800 is fixed with the changes in #801, so this PR is no longer needed.

Issue #755, opened 2021-07-17
🚧 Naming conventions for boolean switches


s-weigand:
For consistencies sake and also easier naming for use, we should come up with some naming convention.
While reviewing #747 I came across some misleading names like `object.full_model` and `object.global_model`, which are just boolean switches.
IMHO going with a pattern like `is_<something>` makes it a lot easier to read the code and quickly see what is going on.

jsnel:
Some work for this is done on the staging branch, is this still an issue? If so the PR which ultimately merges staging back into main should be listed as a Linked pull request to close this issue.

s-weigand:
Not sure if we are consistent across the whole package (`main` + `staging`).

Also to prevent bugs like https://github.com/glotaran/pyglotaran/pull/747#discussion_r670785543

It is a good practice to use identity comparison against the boolean singletons
```python
if is_something() is True:
    ...
```

s-weigand:
This will be done inclement, when ever we are touching code or in a big refactor PR dedicated to this issue.

PR #808, opened 2021-09-11
🩹 Fix coherent artifact crash for index dependent models

### LINKED ISSUE #793: 🐛 CoherentArtifactMegacomplex.finalize_data crashes due to wrong dimensions
### Version information

- pyglotaran version: staging
- Python version: all
- Operating System: all

### Describe the bug

When using a `CoherentArtifactMegacomplex` pyglotaran crashes while creating the result, when trying to assign `coherent_artifact_concentration`

https://github.com/glotaran/pyglotaran/blob/b28179970cd7ab0d9b7e276c3d25ff7dae69e7cd/glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py#L76-L79

This is due to the fact that `dataset.matrix.sel(clp_label=self.compartments()).values` has the dimensions (`global_dimension`, `model_dimension`, `"coherent_artifact_order"`) and we try to create a `DataArray` of dimensions (`model_dimension`, `"coherent_artifact_order"`).

### To Reproduce

Run `pyglotaran_examples/ex_doas_beta/ex_doas_beta.py`from  [this commit](https://github.com/s-weigand/pyglotaran-examples/commit/d4460b6045c6838523249255d436603fd1c96471)

### Expected behavior

Result gets created without crashing.

### Possible Solutions

1. Save all data with the full dimensions (`global_dimension`, `model_dimension`, `"coherent_artifact_order"`)
2. Only use a single `global_dimension` e.g. `dataset.matrix.sel(clp_label=self.compartments()).values[0, :, :]`
3. Use an aggregation function along the `global_dimension` e.g. `dataset.matrix.sel(clp_label=self.compartments()).sum(global_dimension).values` or `dataset.matrix.sel(clp_label=self.compartments()).mean(global_dimension).values`

s-weigand:
For details look at #793

This PR is best reviewed per commit.

### Change summary

- 🩹 Fixes crashing of CoherentArtifactMegacomplex.finalize_data for index dependent models
- ♻️ Refactored test_coherent_artifact to be more readable
- 🧪 Parametrized test_coherent_artifact to cover index dependent case
- ♻️ Renamed `DatasetModel.index_dependent` ->`DatasetModel.is_index_dependent` (see #755)

### Checklist

### Closes issues

closes #793


jsnel (approved):
Code reviewed: :ok_hand:

s-weigand:
Now it should be ready.
I'll add the sourcery configuration commit in a different PR.

glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
```diff
@@ -73,8 +73,11 @@ def finalize_data(
             global_dimension = dataset_model.get_global_dimension()
             model_dimension = dataset_model.get_model_dimension()
             dataset.coords["coherent_artifact_order"] = np.arange(1, self.order + 1)
+            concentration_dimensions = (model_dimension, "coherent_artifact_order")
```
joernweissenborn, line 76: nitpicky, but don't call it "concentration". A coherent artifact is basically backreflections etc introduced by sample container. Call it "coherent_artifact_dimensions" or just "dimensions"

joernweissenborn (changes requested):

glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
s-weigand: The reason I called it that way is because those are the dimensions for `coherent_artifact_concentration` (line 79).
@joernweissenborn  @jsnel Should we maybe rename `coherent_artifact_concentration`?

joernweissenborn: Hmm, this was my mistake naming it like that. "Amplitude" would be the right name. Could change the dataset item from `coherent_artifact_dimension` to `coherent_artifact_amplitude` in this PR?

jsnel: Agree that 'concentration' is not the right terminology here. Amplitudes is a bit ambiguous (could refer to both the amplitude as function of time at a given wavelength, or the (spectral) amplitude (clp) at a given wavelength). 
Maybe coherent_artifact_response is more appropriate here, since that's a commonly used term. See e.g.: 
https://link.springer.com/article/10.1134/1.1884668

joernweissenborn (approved):
lgtm

PR #811, opened 2021-09-11
🔧 Added sourcery config file

s-weigand:
Added sourcery config file and added 'simplify-boolean-comparison' to ignore.
This prevents expressions like `if something is True` to be refactored.

### Background

'if something is True' prevents bugs where something is a callable, where 'if something():' would evaluate to False but 'if something:' evaluated to true since the callable object exists.

While using this explicit syntax creates a longer instruction list, the performance impact is neglectable.

```python
In [1]: from dis import dis

In [2]: def foo():
   ...:     var = True
   ...:     if var is True:
   ...:         return False
   ...:

In [3]: def bar():
   ...:     var = True
   ...:     if var:
   ...:         return False
   ...:

In [4]: dis(foo)
  2           0 LOAD_CONST               1 (True)
              2 STORE_FAST               0 (var)

  3           4 LOAD_FAST                0 (var)
              6 LOAD_CONST               1 (True)
              8 COMPARE_OP               8 (is)
             10 POP_JUMP_IF_FALSE       16

  4          12 LOAD_CONST               2 (False)
             14 RETURN_VALUE
        >>   16 LOAD_CONST               0 (None)
             18 RETURN_VALUE

In [5]: dis(bar)
  2           0 LOAD_CONST               1 (True)
              2 STORE_FAST               0 (var)

  3           4 LOAD_FAST                0 (var)
              6 POP_JUMP_IF_FALSE       12

  4           8 LOAD_CONST               2 (False)
             10 RETURN_VALUE
        >>   12 LOAD_CONST               0 (None)
             14 RETURN_VALUE

In [6]: %timeit foo()
117 ns ± 0.643 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: %timeit bar()
115 ns ± 1.36 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

### Change summary

- Added sourcery config file and added `simplify-boolean-comparison` to ignore

### Checklist

### DIFF
--- a/.sourcery.yaml
+++ b/.sourcery.yaml
@@ -1,0 +1,18 @@
+ignore: []
+
+refactor:
+  skip: [simplify-boolean-comparison]
+
+metrics:
+  quality_threshold: 25.0
+
+clone_detection:
+  min_lines: 3
+  min_duplicates: 2
+  identical_clones_only: false
+
+github:
+  labels: []
+  ignore_labels: [sourcery-ignore]
+  request_review: author
+  sourcery_branch: sourcery/{base_branch}


jsnel (approved):
👌 LGTM

s-weigand:
Just for reference when we need to tweak sourcery again here [are the rules](https://github.com/sourcery-ai/sourcery/wiki/Current-Refactorings)

PR #843, opened 2021-09-30
🩹 Fix parameter expression parsing

s-weigand:
The naive approach using `expression.replace` lead to problems when an expression for a parameter was part of another expression for a parameter.

Example:
The expression:
`"100 - $inputs1.s1 - $inputs1.s3 - $inputs1.s8 - $inputs1.s12"`
should have been transformed to:
`"100 - group.get('inputs1.s1').value - group.get('inputs1.s3').value - group.get('inputs1.s8').value - group.get('inputs1.s12').value"`
but instead was transformed to:
`"100 - group.get('inputs1.s1').value - group.get('inputs1.s3').value - group.get('inputs1.s8').value - group.get('inputs1.s1').value2"`

Since `"inputs1.s1"` partially matches `"inputs1.s12"`

I also moved the regex compiling out of the class, so it only gets executed when the module is loaded.

### Change summary

- 🩹♻️🧪 Fixed Parameter expression transformation
- 👌 Added missing space in exception text of update_parameter_expression

### Checklist

### DIFF
--- a/glotaran/parameter/parameter.py
+++ b/glotaran/parameter/parameter.py
@@ -29,13 +29,14 @@ class Keys:
     VARY = "vary"
 
 
+PARAMETER_EXPRESION_REGEX = re.compile(r"\$(?P<parameter_expression>[\w\d\.]+)((?![\w\d\.]+)|$)")
+"""A regexpression to find and replace parameter names in expressions."""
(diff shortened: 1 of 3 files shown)


joernweissenborn (approved):
lgtm

jsnel (approved):
LGTM. Reviewed and tested out locally. 👍

PR #848, opened 2021-10-03
🩹 Use a context manager when opening a nc dataset

### LINKED ISSUE #847: 🐛 Dataset file locked during lifetime of script
### Describe the bug

A minor issue when reading in / loading data from nc file. When doing so the files are locked throughout the runtime of the script.

### Solution

Instead of 
```python
@register_data_io("nc")
class NetCDFDataIo(DataIoInterface):
    def load_dataset(self, file_name: str) -> xr.Dataset | xr.DataArray:
        return xr.open_dataset(file_name)
```

One can instead use:
```python
@register_data_io("nc")
class NetCDFDataIo(DataIoInterface):
    def load_dataset(self, file_name: str) -> xr.Dataset | xr.DataArray:
        with xr.open_dataset(file_name) as ds:
            return ds.load()
```
which avoids file locking.

jsnel:
avoid file locking when calling xarray.open_dataset

### Checklist

### Closes issues

closes #847

### DIFF
--- a/glotaran/builtin/io/netCDF/netCDF.py
+++ b/glotaran/builtin/io/netCDF/netCDF.py
@@ -11,7 +11,8 @@ from glotaran.project import default_data_filters
 @register_data_io("nc")
 class NetCDFDataIo(DataIoInterface):
     def load_dataset(self, file_name: str) -> xr.Dataset | xr.DataArray:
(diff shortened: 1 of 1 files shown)


s-weigand (approved):
LGTM 👍

PR #869, opened 2021-10-17
✨ Project Class

joernweissenborn:
This PR adds a basic `Project` class.

The `Project` class is intended to serve as a standard API to `pyglotaran`.

Add the moment, using `pyglotaran` as a python package requires alot of imports from different sub packages. This ok because our current API is kind of a "super user" API. The nature of it is that it changes alot, which is not user friendly. On the other hand side, we now have a deprecation framework, which is very useful but slows down development.

The `Project` class mitigates this. It is designed as a user API, escpeicially with having GUIs and stuff in mind. It is meant to be a representation of the actual iterative process of global target analysis. 

### Details

The `Project` class as implemented in this PR is basically a kind of "API aggregator" plus a KISS principle based "database" called `ProjectRegistry`.

It is essentially a collection of a project file plus

* a model folder
* a parameters folder
* a data folder

This folders are represented by  `ProjectRegistry` objects. They work by leveraging the io plugin system and `pathlib`. 

A worwflow in looks like this:

```py
from glotaran.project import Project

EXAMPLE_DIR = "my_project"
EXAMPLE_DATA = "testdata.nc"

Project.create(EXAMPLE_DIR) # create the project

gtap = Project.open(EXAMPLE_DIR) # open a project
gtap.import_data(EXAMPLE_DATA, name="dataset_1")

gtap.generate_model(
    generator_name="simple_model",
    generator="decay_sequential",
    generator_arguments={"nr_compartments": 3, "irf": True},
)

gtap.generate_parameters("simple_model", name="simple_params")

# Note: For now, you need to open the generated models in an editor and change the starting values.

gtap.optimize(
    model="simple_model",
    parameters="simple_params",
    maximum_number_function_evaluations=1,
    name="simple_result",
) # optimize the model parameters

result = gtap.load_result("simple_result") # load the result

result_path = gtap.get_result_path("simple_result") # convenience for e.g. plotting
```
 
### Checklist

### Closes issues

closes #XXXX

### DIFF
--- a/glotaran/model/model.py
+++ b/glotaran/model/model.py
@@ -125,7 +125,7 @@ class Model:
                 warn(f"Unknown model item type '{item_name}'.")
                 continue
 
-            if is_list := isinstance(getattr(model, item_name), list):
+            if isinstance(getattr(model, item_name), list):
                 model._add_list_items(item_name, items)
             else:
                 model._add_dict_items(item_name, items)
@@ -136,7 +136,7 @@ class Model:
 
         for label, item in items.items():
             item_cls = self.model_items[item_name]
-            if is_typed := hasattr(item_cls, "_glotaran_model_item_typed"):
+            if hasattr(item_cls, "_glotaran_model_item_typed"):
                 if "type" not in item and item_cls.get_default_type() is None:
                     raise ValueError(f"Missing type for attribute '{item_name}'")
                 item_type = item.get("type", item_cls.get_default_type())
@@ -153,7 +153,7 @@ class Model:
 
         for item in items:
             item_cls = self.model_items[item_name]
-            if is_typed := hasattr(item_cls, "_glotaran_model_item_typed"):
+            if hasattr(item_cls, "_glotaran_model_item_typed"):
                 if "type" not in item:
                     raise ValueError(f"Missing type for attribute '{item_name}'")
                 item_type = item["type"]
--- a/glotaran/model/model.py
+++ b/glotaran/model/model.py
@@ -312,14 +312,14 @@ class Model:
 
         return model_dict
 
-    def get_parameters(self) -> list[str]:
-        parameters = []
+    def get_parameter_labels(self) -> list[str]:
+        parameter_labels = []
         for item_name in self.model_items:
             items = getattr(self, item_name)
             item_iterator = items if isinstance(items, list) else items.values()
             for item in item_iterator:
-                parameters += item.get_parameters()
-        return parameters
+                parameter_labels += item.get_parameter_labels()
+        return parameter_labels
 
     def need_index_dependent(self) -> bool:
         """Returns true if e.g. clp_relations with intervals are present."""
--- a/glotaran/model/model.py
+++ b/glotaran/model/model.py
@@ -125,9 +125,7 @@ class Model:
                 warn(f"Unknown model item type '{item_name}'.")
                 continue
 
-            is_list = isinstance(getattr(model, item_name), list)
-
-            if is_list:
+            if is_list := isinstance(getattr(model, item_name), list):
                 model._add_list_items(item_name, items)
             else:
                 model._add_dict_items(item_name, items)
@@ -138,8 +136,7 @@ class Model:
 
         for label, item in items.items():
             item_cls = self.model_items[item_name]
-            is_typed = hasattr(item_cls, "_glotaran_model_item_typed")
-            if is_typed:
+            if is_typed := hasattr(item_cls, "_glotaran_model_item_typed"):
                 if "type" not in item and item_cls.get_default_type() is None:
                     raise ValueError(f"Missing type for attribute '{item_name}'")
                 item_type = item.get("type", item_cls.get_default_type())
@@ -156,8 +153,7 @@ class Model:
 
         for item in items:
             item_cls = self.model_items[item_name]
-            is_typed = hasattr(item_cls, "_glotaran_model_item_typed")
-            if is_typed:
+            if is_typed := hasattr(item_cls, "_glotaran_model_item_typed"):
                 if "type" not in item:
                     raise ValueError(f"Missing type for attribute '{item_name}'")
                 item_type = item["type"]
@@ -374,8 +370,7 @@ class Model:
         """
         result = ""
 
-        problems = self.problem_list(parameters)
-        if problems:
+        if problems := self.problem_list(parameters):
             result = f"Your model has {len(problems)} problems:\n"
             for p in problems:
                 result += f"\n * {p}"
--- a/glotaran/model/model.py
+++ b/glotaran/model/model.py
@@ -212,14 +212,14 @@ class Model:
     def _add_dataset_property(self, property_name: str, dataset_property: dict[str, any]):
         if property_name in self._dataset_properties:
             known_type = (
-                self._dataset_properties[property_name]
-                if not isinstance(self._dataset_properties, dict)
-                else self._dataset_properties[property_name]["type"]
+                self._dataset_properties[property_name]["type"]
+                if isinstance(self._dataset_properties, dict)
+                else self._dataset_properties[property_name]
             )
             new_type = (
-                dataset_property
-                if not isinstance(dataset_property, dict)
-                else dataset_property["type"]
+                dataset_property["type"]
+                if isinstance(dataset_property, dict)
+                else dataset_property
             )
             if known_type != new_type:
                 raise ModelError(
--- a/glotaran/model/model.py
+++ b/glotaran/model/model.py
@@ -42,6 +42,10 @@ default_dataset_properties = {
     "scale": {"type": Parameter, "default": None, "allow_none": True},
 }
 
+root_parameter_error = ModelError(
+    "The root parameter group cannot contain both groups and parameters."
+)
+
 
 class Model:
     """A base class for global analysis models."""
@@ -316,6 +320,32 @@ class Model:
             for item in item_iterator:
                 parameter_labels += item.get_parameter_labels()
         return parameter_labels
+
+    def generate_parameters(self) -> dict | list:
+        parameters: dict | list = {}
+        for parameter in self.get_parameter_labels():
+            groups = parameter.split(".")
+            label = groups.pop()
+            if len(groups) == 0:
+                if isinstance(parameters, dict):
+                    if len(parameters) != 0:
+                        raise root_parameter_error
+                    else:
+                        parameters = []
+                parameters.append(Parameter.create_default_list(label))
+            else:
+                if isinstance(parameters, list):
+                    raise root_parameter_error
+                this_group = groups.pop()
+                group = parameters
+                for name in groups:
+                    if name not in group:
+                        group[name] = {}
+                    group = group[name]
+                if this_group not in group:
+                    group[this_group] = []
+                group[this_group].append(Parameter.create_default_list(label))
+        return parameters
 
     def need_index_dependent(self) -> bool:
         """Returns true if e.g. clp_relations with intervals are present."""
--- a/glotaran/project/project.py
+++ b/glotaran/project/project.py
@@ -115,7 +115,10 @@ class Project:
         project_dict = load_dict(file, True)
         project_dict["file"] = file
         project_dict["folder"] = folder
-        return cls(**project_dict)
+        version = project_dict.pop("version")
+        project = cls(**project_dict)
+        project.version = version
+        return project
 
     @property
     def has_data(self) -> bool:
--- a/glotaran/project/project.py
+++ b/glotaran/project/project.py
@@ -53,7 +53,7 @@ class Project:
         self._result_registry = ProjectResultRegistry(self.folder)
 
     @staticmethod
-    def create(folder: str | Path):
+    def create(folder: str | Path, overwrite: bool = False):
         """Create a new project folder and file.
 
         Parameters
@@ -61,14 +61,24 @@ class Project:
         folder : str | Path | None
             The folder where the project will be created. If ``None``, the current work
             directory will be used.
+        overwrite: bool
+            Whether to overwrite an existing project file.
+
+        Raises
+        ------
+        FileExistsError
+            Raised if the project file already exists and `overwrite=False`.
         """
         from glotaran import __version__ as gta_version
 
         project_folder = make_path_absolute_if_relative(Path(folder))
         project_folder.mkdir(parents=True, exist_ok=True)
         project_file = project_folder / PROJECT_FILE_NAME
-        with open(project_file, "w") as f:
-            f.write(TEMPLATE.format(gta_version=gta_version))
+        if project_file.exists() and not overwrite:
+            raise FileExistsError(
+                f"Project file '{project_file}' already exist. Set `overwrite=True` to overwrite."
+            )
+        project_file.write_text(TEMPLATE.format(gta_version=gta_version))
 
     @classmethod
     def open(cls, project_folder_or_file: str | Path, create_if_not_exist: bool = True):
--- a/glotaran/project/project.py
+++ b/glotaran/project/project.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import re
 from dataclasses import dataclass
 from dataclasses import field
+from importlib.metadata import distribution
 from pathlib import Path
 from textwrap import dedent
 from typing import Any
@@ -46,6 +47,7 @@ class Project:
         if self.folder is None:
             self.folder = self.file.parent
         self.folder = Path(self.folder)
+        self.version = distribution("pyglotaran").version
 
         self._data_registry = ProjectDataRegistry(self.folder)
         self._model_registry = ProjectModelRegistry(self.folder)
@@ -74,8 +76,6 @@ class Project:
         FileExistsError
             Raised if the project file already exists and `allow_overwrite=False`.
         """
-        from glotaran import __version__ as gta_version
-
         project_folder = make_path_absolute_if_relative(Path(folder))
         project_folder.mkdir(parents=True, exist_ok=True)
         project_file = project_folder / PROJECT_FILE_NAME
@@ -84,8 +84,8 @@ class Project:
                 f"Project file '{project_file}' already exist. "
                 "Set `allow_overwrite=True` to overwrite."
             )
-        project_file.write_text(TEMPLATE.format(gta_version=gta_version))
-        return Project.open(project_file)
+        project_file.write_text(TEMPLATE.format(gta_version=distribution("pyglotaran").version))
+        return Project.open(project_file, create_if_not_exist=True)
 
     @classmethod
     def open(cls, project_folder_or_file: str | Path, create_if_not_exist: bool = True) -> Project:
@@ -114,9 +114,9 @@ class Project:
         else:
             file = folder / PROJECT_FILE_NAME
 
-        if not file.exists():
-            if not create_if_not_exist:
-                raise FileNotFoundError(f"Project file {file} does not exists.")
+        if file.is_file() is False:
+            if create_if_not_exist is False:
+                raise FileNotFoundError(f"Project file {file.as_posix()} does not exists.")
             Project.create(folder)
 
         project_dict = load_dict(file, True)
@@ -139,12 +139,12 @@ class Project:
         return not self._data_registry.empty
 
     @property
-    def data(self) -> dict[str, str]:
+    def data(self) -> dict[str, Path]:
         """Get all project datasets.
 
         Returns
         -------
-        dict[str, str]
+        dict[str, Path]
             The models of the datasets.
         """
         return self._data_registry.items
@@ -170,7 +170,7 @@ class Project:
         try:
             return self._data_registry.load_item(name)
         except ValueError as e:
-            raise ValueError(f"Dataset '{name}' does not exist.") from e
+            raise ValueError(f"Dataset {name!r} does not exist.") from e
 
     def import_data(
         self,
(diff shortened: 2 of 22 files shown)


jsnel:
After rebasing, please sit together @s-weigand and @joernweissenborn and try to turn one or two of the examples into a project. e.g. [this one](https://github.com/glotaran/pyglotaran-examples/tree/main/pyglotaran_examples/ex_two_datasets)

glotaran/project/project_parameter_registry.py
```diff
@@ -0,0 +1,116 @@
+"""The glotaran parameter registry module."""
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Literal
+
+from yaml import dump
```
s-weigand, line 7: `yaml` isn't a builtin module and we changed for `pyyaml` to `ruamel.yaml`  use this instead so the benchmarks won't crash. This only works in the other tests because we get `pyyaml` as an indirect dependency from one of our test dependencies.

glotaran/project/project.py
```diff
@@ -0,0 +1,444 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from os import getcwd
+from pathlib import Path
+from typing import Any
+from typing import Literal
+
+import xarray as xr
+from yaml import safe_load
```
s-weigand, line 11: `yaml` isn't a builtin module and we changed for `pyyaml` to `ruamel.yaml`  use this instead so the benchmarks won't crash. This only works in the other tests because we get `pyyaml` as an indirect dependency from one of our test dependencies.
`ruamel.yaml` also has a [save loader](https://yaml.readthedocs.io/en/latest/basicuse.html#basic-usage)

joernweissenborn: I resolved from using built in yml and expose the load function.

glotaran/project/project.py
```diff
@@ -0,0 +1,446 @@
...
+            project_folder.mkdir()
+        name = name or project_folder.name
+        project_file = project_folder / PROJECT_FILE_NAME
+        with open(project_file, "w") as f:
+            f.write(TEMPLATE.format(gta_version=gta_version, name=name))
+
+        return cls.open(project_file)
+
+    @classmethod
+    def open(cls, project_folder_or_file: str | Path):
```
s-weigand, line 89: This will have a problem when using relative paths and executing from different paths.
We [already solved](https://github.com/glotaran/pyglotaran-extras/blob/f4b64beb621195d428e1bfd70c819af026889a38/pyglotaran_extras/io/setup_case_study.py#L52) this problem in the extras in a way that it works for scripts and notebooks.
Just copy that function over to `glotaran.utils.io`

glotaran/project/project.py
```diff
@@ -0,0 +1,446 @@
...
+        if isinstance(self.folder, str):
+            self.folder = Path(self.folder)
+
+        self._data_registry = ProjectDataRegistry(self.folder)
+        self._model_registry = ProjectModelRegistry(self.folder)
+        self._parameter_registry = ProjectParameterRegistry(self.folder)
+        self._result_registry = ProjectResultRegistry(self.folder)
+
+    @classmethod
+    def create(cls, name: str | None = None, folder: str | Path | None = None) -> Project:
```
s-weigand, line 58: Same as for `open`:

> This will have a problem when using relative paths and executing from different paths.
We [already solved](https://github.com/glotaran/pyglotaran-extras/blob/f4b64beb621195d428e1bfd70c819af026889a38/pyglotaran_extras/io/setup_case_study.py#L52) this problem in the extras in a way that it works for scripts and notebooks.
Just copy that function over to `glotaran.utils.io`

glotaran/project/project.py
```diff
@@ -0,0 +1,446 @@
...
+        ------
+        ValueError
+            Raised if the dataset does not exist.
+        """
+        try:
+            return self._data_registry.load_item(name)
+        except ValueError:
+            raise ValueError(f"Dataset '{name}' does not exist.")
+
+    def import_data(self, path: str | Path, name: str | None = None):
```
s-weigand, line 158: Please allow use of [`DatasetMappable`](https://github.com/glotaran/pyglotaran/blob/a7f67d4ad7ec5b45972c5aca3163e234abdc4db0/glotaran/typing/types.py#L11) and use [`load_datasets`](https://github.com/glotaran/pyglotaran/blob/a7f67d4ad7ec5b45972c5aca3163e234abdc4db0/glotaran/utils/io.py#L166) under the hood. You could also use `load_datasets` and `Path.glob` in `load_data`.

s-weigand:
Ok so the `yaml` issue using `offset=2` is as follows:

We read something like this (in python `list[dict[str, float | list[float | str]]]`):
```yaml
weights:
  - datasets: [ds1]
    global_interval: [520, 999]
    value: 0.4
```

And after optimizing  we serialize the `Model` to a `dict` and dump it to `yaml`

Which results in this:
```yaml
weights:
  - datasets: [ds1]
  global_interval: [520, 999]
  value: 0.4
```
Which is not what we started with and is **broken syntax**, causing an error.
This error was **NOT** caught by the tests, but when using this code in a case study where `project.load_result` was used.
@joernweissenborn Please add tests for `project.load_result` (you can run `pytest --cov=glotaran --cov-config pyproject.toml --cov-report html --cov-report term glotaran` to easily spot which other parts also need more test coverage, just open `htmlcov/index.html`)

The current "fix" is to use `offset=0` which when using `sequence=2` is the [encouraged style by the author to `ruamel.yaml`](https://yaml.readthedocs.io/en/latest/detail.html#indentation-of-block-sequences)

This results in the following yaml:
```yaml
weights:
- datasets: [Krg_grga]
  global_interval: [520, 535]
  value: 0.4
```

While this is valid syntax, it breaks the flow of having child elements two spaces indented compared to their parent label.

In the end this is an upstream bug at `ruamel.yaml`, we might be able to fix it in our case when implementing roundtrip saving (the problem is that we will need to replace each value in the nested `CommentedMap` structure recursively since replacing it with a dict will destroy the roundtripping and there is no nice API to do that yet) .
We will need roundtripping for `yaml` in any case since the `model.yml` in in the results folder has a different order of elements that the initial model, which creates a lot of mental load for the user (ziczac comparing different results with the current model).
And exporting a result with all comments missing would also be an issue.

@jsnel For this PR I would say we change the tests to comply with the `offset=0` "fix", since this is a writing only issue only and reading works fine as the passing integration tests show.

Just as a side note and reference if we can't solve this issue by roundtripping support we could also have a look at the community fork of `ruamel.yaml` which is called [`ruyaml`](https://github.com/pycontribs/ruyaml) and on github, while `ruamel.yaml` is on [SourceForge](https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/) (which IMHO is a horrible SCM 🤢)

jsnel:
@s-weigand Let's go with the one working solution right now:
> The current "fix" is to use offset=0 which when using sequence=2 is the encouraged style by the author to ruamel.yaml

jsnel:
The syntax in the PR description is invalid, please revise. Could be more along the lines of:

```python
# %%
from pathlib import Path
from glotaran.project import Project

EXAMPLE_DIR = Path(__file__).parent
DATA_PATH1 = "data/original_dataset1.ascii"
DATA_PATH2 = "data/original_dataset2.ascii"

gtap = Project(
    version="2", folder=EXAMPLE_DIR / "project", file=EXAMPLE_DIR / "simple_project.gta"
) # version seems to be unused

# currently only works if you use dataset_1 etc as lables, not dataset1 or my_datasetA, my_datasetB
gtap.import_data(EXAMPLE_DIR / DATA_PATH1, name="dataset_1")
gtap.import_data(EXAMPLE_DIR / DATA_PATH2, name="dataset_2")

gtap.generate_model(
    name="simple_model",
    generator="decay_sequential",
    generator_arguments={"nr_compartments": 3, "irf": True},
)

gtap.generate_parameters("simple_model", name="simple_params")

gtap.optimize(
    model="simple_model",
    parameters="simple_params",
    maximum_number_function_evaluations=1,
    name="simple_result",
)

```

jsnel (changes requested):
Please update the issues description and take into account comments shared in Slack.

https://github.com/glotaran/pyglotaran/pull/869#issuecomment-1042627298

glotaran/project/project_result_registry.py
```diff
@@ -0,0 +1,75 @@
...
+        -------
+        str :
+            A result name.
+        """
+        i = 0
+        while True:
+            result_name = f"{model_name}_run_{i}"
+            if not (self.directory / result_name).exists():
+                return result_name
+            i += 1
```
s-weigand, line 62: If a user were to inspect their results (let's say they have 10) and clean up some experimental ones (e.g. 2-4) the results would get out of order which makes it really hard to keep track of what happened when especially without a log.
Thus I would recommend using `self.directory.glob(f"{model_name}_run_*")`, and using a regex to find the highest count.

glotaran/project/test/test_project.py
```diff
@@ -0,0 +1,168 @@
...
+    assert not project.has_parameters
+    assert not project.has_results
+
+
+def test_generate_model(project_folder, project_file):
+    project = Project.open(project_file)
+
+    project.generate_model("test_model", "decay_parallel", {"nr_compartments": 5})
+
+    model_folder = Path(project_folder) / "models"
```
s-weigand, line 57: Redundant `Path` initialization since `project_folder` always is a `Path` (See other copypasta occurrences)

s-weigand:
@jsnel good catch
> ```python
> # currently only works if you use dataset_1 etc as lables, not dataset1 or my_datasetA, my_datasetB
> ```
Looks like the generator only work with a single dataset named `dataset_1` since this is hardcoded and there are no options.
https://github.com/glotaran/pyglotaran/blob/f1413eeb66787dc73916ec2fbb13741856aa7311/glotaran/project/generators/generator.py#L44
This means we can consider this functionality as broken.

glotaran/project/test/test_project.py
```diff
@@ -0,0 +1,168 @@
...
+    model_file.write_text(example_model_yml)
+
+    parameters_file = project_folder / "parameters" / "sequential.csv"
+    save_parameters(example_parameter, parameters_file)
+
+    data_folder = project_folder / "data"
+    assert data_folder.exists()
+    data_file = data_folder / "dataset_1.nc"
+    data_file.unlink()
+    save_dataset(example_dataset, data_file)
```
s-weigand, line 142: Why do we need to delete and then save files over and over again?
I guess rewriting the fixtures with `tmp_path` would save us some headaches also, especially with parametrized tests it makes debugging easier since the parameters get appended to the folder name (here something like `test_run_optimization-test_run` and `test_run_optimization-None`)

s-weigand (changes requested):
@joernweissenborn Sorry for the long delay, but now you finally get my review 😅
Currently, there are some nitpicky issues with the code, and as @jsnel  also pointed out usability issues that need to be addressed before we can merge this.

glotaran/project/project_model_registry.py
```diff
@@ -0,0 +1,45 @@
...
+        Parameters
+        ----------
+        name : str
+            The name of the model.
+        generator : str
+            The generator for the model.
+        generator_arguments : GeneratorArguments
+            Arguments for the generator.
+        """
+        model_yml = generate_model_yml(
```
s-weigand, line 42: Make writing a model file conditional and add an additional parameter `overwrite_existing`
```python
if (self._directory / f"{name}.yml").is_file() is not True or overwrite_existing is True:
```

glotaran/project/project.py
joernweissenborn: I do not think that this should be done.

The `import_dataset` mapping is only used once per project, thus this convinience has very little gain for users with the trade off of making the method sequence more complicated.

Another reason is that in future there might be more option to this like options to create an SVD or not and other preprocessing which should and must be set per dataset.

Last but not least: for users which just want to add a bulk of dataset can still simply copy them to the projects data folder. This is also the way to go if a user prefers to have e.g. the dataset as ascii wheres importing always converts to nc.

If it turns out that this functionality is really needed, we can add a method `import_multiple_data` in a different PR.

glotaran/model/model.py
```diff
@@ -312,13 +312,39 @@ def as_dict(self) -> dict:
...
         for item_name in self.model_items:
             items = getattr(self, item_name)
             item_iterator = items if isinstance(items, list) else items.values()
             for item in item_iterator:
-                parameters += item.get_parameters()
+                parameter_labels += item.get_parameter_labels()
+        return parameter_labels
+
+    def generate_parameters(self) -> dict | list:
+        parameters: dict | list = {}
```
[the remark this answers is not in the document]
joernweissenborn: A unit test has been added.

The typing I cannot fix now since I don't now how to create types referencing themselves. I created #1020 to work on this in a future PR.

glotaran/project/project.py
```diff
@@ -0,0 +1,468 @@
...
+        ----------
+        folder : str | Path | None
+            The folder where the project will be created. If ``None``, the current work
+            directory will be used.
+        """
+        from glotaran import __version__ as gta_version
+
+        project_folder = make_path_absolute_if_relative(Path(folder))
+        if not project_folder.exists():
+            project_folder.mkdir()
```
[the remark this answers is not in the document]
joernweissenborn: Done

glotaran/project/project.py
```diff
@@ -0,0 +1,468 @@
...
+        project_file = project_folder / PROJECT_FILE_NAME
+        with open(project_file, "w") as f:
+            f.write(TEMPLATE.format(gta_version=gta_version))
```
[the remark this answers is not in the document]
joernweissenborn: Done, overwrite must now be explicitly set, otherwise `FileExistsError` is raised.

glotaran/project/project.py
```diff
@@ -0,0 +1,468 @@
...
+        """
+        try:
+            return self._model_registry.load_item(name)
+        except ValueError:
+            raise ValueError(f"Model '{name}' does not exist.")
+
+    def generate_model(
+        self,
+        name: str,
+        generator: str,
```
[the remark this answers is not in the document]
joernweissenborn: Done.

About instances: This would be rather useless. If a user has his own generator which is not available via `generator_name`, then she should be able write mode files herself anyway. I added convenience function to get a project's model and data directory to aid with that. The other way would be to register her generator as plugin, but for this we need a plugin system for generators which is way out of this PR's scope.

glotaran/project/project_data_registry.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        if not isinstance(path, Path):
+            path = Path(path)
+
+        name = name or path.with_suffix("").name
+        data_path = self.directory / f"{name}.nc"
+
+        dataset = load_dataset(path)
+        dataset.attrs.pop("loader")
+        dataset.to_netcdf(data_path)
```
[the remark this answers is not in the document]
joernweissenborn: done

glotaran/project/test/test_project.py
joernweissenborn: Well, the simple thing is that I want to test optimization and I use the example sequential model. The deletion is because the test is parameterized. There can be exists checks added, new fixtures made or whatever, but it all adds complexity in it's own way and choose the KISS way.

NOTE: you cannot do this prior to tests, otherwise e.g. the initial `assert not project has_models` would fail.

If you find a more easy way you can implement it, I had no trouble debugging and I don't see where one would have troubles. Thus I leave it like this.

glotaran/project/project_model_registry.py
joernweissenborn: done

glotaran/project/project_parameter_registry.py
```diff
@@ -0,0 +1,55 @@
...
+        ----------
+        model : Model
+            The model.
+        name : str | None
+             The name of the parameters.
+        fmt : Literal["yml", "yaml", "csv"]
+            The parameter format.
+        """
+        parameters = model.generate_parameters()
+        parameter_file = self.directory / f"{name}.{fmt}"
```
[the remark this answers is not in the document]
joernweissenborn: done

glotaran/project/project.py
```diff
@@ -0,0 +1,508 @@
...
+from glotaran.project.project_model_registry import ProjectModelRegistry
+from glotaran.project.project_parameter_registry import ProjectParameterRegistry
+from glotaran.project.project_result_registry import ProjectResultRegistry
+from glotaran.project.result import Result
+from glotaran.project.scheme import Scheme
+from glotaran.utils.io import make_path_absolute_if_relative
+from glotaran.utils.ipython import MarkdownStr
+
+TEMPLATE = """version: {gta_version}
+"""
```
s-weigand, line 26: One more nitpick 😅
```diff
- TEMPLATE = """version: {gta_version}
- """
+ TEMPLATE = "version: {gta_version}"
```
Also since it is only used in one place it wouldn't need to be a global variable.

s-weigand (changes requested):
Some more nitpicks sorry for that 🙈

glotaran/project/project_data_registry.py
```diff
@@ -0,0 +1,43 @@
...
+        name : str | None
+            The name of the dataset.
+        """
+        path = Path(path)
+
+        name = name or path.with_suffix("").name
+        data_path = self.directory / f"{name}.nc"
+
+        dataset = load_dataset(path)
+        save_dataset(dataset, data_path)
```
s-weigand, line 43: Again something for `.stem`

```diff
- name = name or path.with_suffix("").name
+ name = name or path.stem
```

The other thing we talked about is that this also needs an `allow_overwrite` argument as protection, so users don't overwrite data accidentally (`UserWarning`).
And the other thing we talked about is that there should be an `ignore_existing` kwarg that defaults to `False` and can be used to silence the warning.

s-weigand:
@jörn as discussed for consistency reasons please rename `overwrite` -> `allow_overwrite`.

glotaran/project/project_data_registry.py
joernweissenborn: Done

glotaran/project/project.py
joernweissenborn: Done. As discussed, in future there will be more added so we leave it as global.

glotaran/project/project_result_registry.py
```diff
@@ -0,0 +1,76 @@
...
+        str :
+            A result name.
+        """
+        previous_results = list(self.directory.glob(f"{model_name}_run_*"))
+        if not previous_results:
+            return f"{model_name}_run_0"
+        previous_results.sort()
+        latest_result = previous_results[-1].with_suffix("").name
+        latest_result_run_nr = int(latest_result.replace(f"{model_name}_run_", ""))
+        return f"{model_name}_run_{latest_result_run_nr+1}"
```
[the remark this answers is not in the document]
joernweissenborn: done

s-weigand (changes requested):
Just tried it out in binder and there are two final nitpicks about useability.

First I saved the dataset to the file that gets imported
```python
save_dataset(dataset, "testdata.nc")
```
And then ran the demo workflow from the description

1. `generate_parameters` and `generate_model` should also have an `ignore_existing` option, like `import_data` (maybe make it a project wide option? 🤔).
2. `load_result` and `get_result_path` should pick up the latest run

```python
result = gtap.load_result("simple_result_run_0") # load the result

result_path = gtap.get_result_path("simple_result_run_0") # convenience for e.g. plotting
```

When this is fixed it is ready to merge from my side.

glotaran/project/test/test_project.py
s-weigand: Let's leave this for a later refactor maybe for now it is fine.

s-weigand:
As discussed with @joernweissenborn I will take care of the changes requested in https://github.com/glotaran/pyglotaran/pull/869#pullrequestreview-922481971

s-weigand:
Took a while but I fixed all issues I still had with this PR (we will leave the test poisoning for a different PR).

@jsnel for a quick usability test in [binder](https://mybinder.org/v2/gh/joernweissenborn/pyglotaran/feature/project/class?urlpath=lab/tree/docs/source/notebooks)
You can use the following:

Save the dataset from the testing data to the file which gets imported
```python
from glotaran.io import save_dataset
from glotaran.testing.simulated_data.sequential_spectral_decay import (
    DATASET as dummy_data,
)

save_dataset(dummy_data, "testdata.nc", allow_overwrite=True)
```

Create project skeleton
```python
from glotaran.project import Project

EXAMPLE_DIR = "my_project"
EXAMPLE_DATA = "testdata.nc"

gtap = Project.open(EXAMPLE_DIR)  # open a project
gtap.import_data(EXAMPLE_DATA, name="dataset_1", ignore_existing=True)

gtap.generate_model(
    model_name="simple_model",
    generator_name="decay_sequential",
    generator_arguments={"nr_compartments": 3, "irf": True},
    ignore_existing=True,
)

gtap.generate_parameters("simple_model", parameters_name="simple_params", ignore_existing=True)
```

Edit the parameters
```csv
label,value,expression,minimum,maximum,non-negative,vary,standard-error
irf.center,0.3,None,,,False,True,None
irf.width,0.1,None,,,False,True,None
rates.species_1,0.5,None,,,False,True,None
rates.species_2,0.3,None,,,False,True,None
rates.species_3,0.1,None,,,False,True,None
```

Optimize the model
```python
# Note: For now, you need to open the generated models in an editor and change the starting values.

gtap.optimize(
    model_name="simple_model",
    parameters_name="simple_params",
    maximum_number_function_evaluations=1,
    result_name="simple_result",
)  # optimize the model parameters

result = gtap.load_result("simple_result")  # load the result

result_path = gtap.get_result_path("simple_result")  # convenience for e.g. plotting
```
@joernweissenborn I also fixed some naming consistency issues please also check my changes.

joernweissenborn:
lgtm

glotaran/project/project.py
jsnel: Noted. Marking as resolved with a note to take these remarks into consideration in a future PR.

PR #893, opened 2021-11-07
✨ Use ruaml.yaml parser for roundtrip support 

### LINKED ISSUE #887: ✨ Use ruaml yaml parser for roundtrip support
### Is your feature request related to a problem? Please describe.

Currently, we are using [`PyYAML`](https://github.com/yaml/pyyaml) as yaml parser, which doesn't support roundtripping of files (removes comment, doesn't honor original order).
Those limitations prevent us from having iterative interactions with model and parameters, which can then be dumped back to file in a user friendly fashion.

### Describe the solution you'd like

If we were to use [`ruaml`](https://yaml.readthedocs.io/en/latest/example.html) instead we could have the much needed roundtrip support, opening new possibilities. We could even programmatically add comments to the user files, e.g. add commented out parameters to the parameter file which are used in a model but aren't defined in the parameter file.

s-weigand:
This PR replaces `pyyaml` with `ruaml.yaml` which allows roundtrip support.

We will need some extra code to update existing yaml files from dicts, but we already get some nice out of the box improvements compared to `pyyaml`:

- Respecting order of dicts (see changes in `test_save_model.py` and `test_save_scheme.py ` )
- [correct parsing of floating point scalars with exponentials](https://yaml.readthedocs.io/en/latest/pyyaml.html#defaulting-to-yaml-1-2-support) (which we currently do manually and should keep for other plugins)
https://github.com/glotaran/pyglotaran/blob/efc333a05b96c67f1043577ab1e7b0eb3bc64a20/glotaran/utils/sanitize.py#L194-L195
- Ability to [add and replace comments](https://yaml.readthedocs.io/en/latest/detail.html#adding-replacing-comments)
  In the future, this would allow us to annotate files with comments e.g. when calling `model.validate(parameters)`.

### Change summary

- ✨ Replaced pyyaml with ruaml.yaml 
- 👌 Set yaml repr for None to null, instead of missing value ('')
- 👌 Changed list element indentation to be consistent with current style
- 🧹 Sorted dependencies in requirements_dev.txt same as setup.cfg 

### Checklist

### Closes issues

closes #887


jsnel (approved):
LGTM! 👍

joernweissenborn:
@s-weigand While we are at it, isn't there maybe cleaner way for parsing tuples then the current with ruaml?

s-weigand:
> @s-weigand While we are at it, isn't there maybe cleaner way for parsing tuples then the current with ruaml?

There might be! This is basically just a dependency swap as preparation for the `0.5.0` release, we will profit from it in `0.6.0` and afterward, but if our users decide that they want to use `main` they at least already have the proper dependencies installed.

PR #907, opened 2021-11-17
🩹 Fixed artifact download in pr_benchmark_reaction workflow

s-weigand:
This change was needed due to updating to actions/github-script@v5 (#845, merged via #884 )

### Change summary

- 🩹 Fixed artifact download in pr_benchmark_reaction workflow 

### DIFF
--- a/.github/workflows/pr_benchmark_reaction.yml
+++ b/.github/workflows/pr_benchmark_reaction.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/github-script@v5
         with:
           script: |
-            var artifacts = await github.actions.listWorkflowRunArtifacts({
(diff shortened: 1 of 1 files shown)


s-weigand:
It doesn't work yet since the workflow which gets executed needs to live on the default branch.

s-weigand:
I switched the default branch on my fork to this branch and made a dummy PR.
[The workflow](https://github.com/s-weigand/pyglotaran/runs/4235753785?check_suite_focus=true) now fails on commenting since this is hardcoded to `glotaran/pyglotaran` and the PR with that number doesn't exist here.
So it works as expected. 🎉

jsnel (approved):
LGTM!

PR #910, opened 2021-11-20
Serveral refactorings in `glotaran.Parameter`

### LINKED ISSUE #902: 🐛 Optimized parameters are missing StdErr when saved to file
### Version information

- pyglotaran version: All
- Python version: All
- Operating System: All

### Describe the bug

The optimized parameters from an analysis when saved to file are missing the StdErr, thus making the result incomplete.

### To Reproduce
Run any example from the [examples repo](https://github.com/glotaran/pyglotaran-examples) or the quickstart.

### Expected behavior

The StdErr of optimized parameters is saved in the parameter file.

### Screenshots

Example from the quickstart.
grafik

grafik

joernweissenborn:
This PR streamlines the to and from dataframe methods of parameter group and ensures all parameter properties are included in the dataframe.

### Change summary

- Added methods `as_dict` and `from_dict` to `Parameter`
- Added methods `to_parameter_dict_list` and `from_parameter_dict_list` to `ParameterGroup`
- Refactored `to_dataframe` and `from_dataframe` to use new methods
- Refactored parameter tests
- Load standard error from df
- save standard error to parameter df
  - `to_datafame` has a switch `as_optimized` now which defaults to true, a later pr must adapt the io API
- Added `Parameter.markdown` as basis for follow up PR
  - This NOT used `ParameterGroup` to markdown. It is intended to be used by `Model.markdown`
  
### Checklist

### Closes issues

closes #902
u

### DIFF
--- a/glotaran/parameter/parameter.py
+++ b/glotaran/parameter/parameter.py
@@ -491,19 +491,18 @@ class Parameter(_SupportsArray):
                 initial_value = initial_parameter.get(self.full_label).value
                 value += f", initial: {initial_value:.2e}"
             md += f"({value})"
+        elif self.expression is not None:
+            expression = self.expression
+            if all_parameter is not None:
+                for match in PARAMETER_EXPRESION_REGEX.findall(expression):
+                    label = match[0]
+                    parameter = all_parameter.get(label)
+                    expression = expression.replace(
+                        "$" + label, f"_{parameter.markdown(all_parameter=all_parameter)}_"
+                    )
+            md += f"({value}={expression})"
         else:
-            if self.expression is not None:
-                expression = self.expression
-                if all_parameter is not None:
-                    for match in PARAMETER_EXPRESION_REGEX.findall(expression):
-                        label = match[0]
-                        parameter = all_parameter.get(label)
-                        expression = expression.replace(
-                            "$" + label, f"_{parameter.markdown(all_parameter=all_parameter)}_"
-                        )
-                md += f"({value}={expression})"
-            else:
-                md += f"({value}, fixed)"
+            md += f"({value}, fixed)"
 
         return MarkdownStr(md)
 
--- a/glotaran/parameter/parameter.py
+++ b/glotaran/parameter/parameter.py
@@ -273,7 +273,7 @@ class Parameter(_SupportsArray):
 
     @property
     def non_negative(self) -> bool:
-        r"""Indicate if the parameter is non-negativ.
+        r"""Indicate if the parameter is non-negative.
 
         If true, the parameter will be transformed with :math:`p' = \log{p}` and
         :math:`p = \exp{p'}`.
@@ -285,7 +285,7 @@ class Parameter(_SupportsArray):
         Returns
         -------
         bool
-            Whether the parameter is non-negativ.
+            Whether the parameter is non-negative.
         """
         return self._non_negative if self.expression is None else False
 
--- a/glotaran/parameter/parameter.py
+++ b/glotaran/parameter/parameter.py
@@ -172,6 +172,8 @@ class Parameter(_SupportsArray):
     def as_dict(self, as_optimized: bool = True) -> dict[str, Any]:
         """Create a dictionary containing the parameter properties.
 
+        Note:
+        -----
         Intended for internal use.
 
         Parameters
--- a/glotaran/parameter/parameter.py
+++ b/glotaran/parameter/parameter.py
@@ -165,7 +165,7 @@ class Parameter(_SupportsArray):
         parameter_dict["label"] = parameter_dict["label"].split(".")[-1]
         if "non-negative" in parameter_dict:
             parameter_dict["non_negative"] = parameter_dict.pop("non-negative")
-        parameter_dict.pop("standart-error", None)
+        parameter_dict.pop("standard-error", None)
 
         return cls(**parameter_dict)
 
@@ -182,7 +182,7 @@ class Parameter(_SupportsArray):
         return {
             "label": self.full_label,
             "value": self.value,
-            "standart-error": self.standard_error,
+            "standard-error": self.standard_error,
             "expression": self.expression,
             "minimum": self.minimum,
             "maximum": self.maximum,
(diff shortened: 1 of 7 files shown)


glotaran/parameter/parameter_group.py
```diff
@@ -239,32 +232,69 @@ def root_group(self) -> ParameterGroup | None:
...
-            parameter_dict["value"].append(parameter.value)
-            parameter_dict["minimum"].append(parameter.minimum)
-            parameter_dict["maximum"].append(parameter.maximum)
-            parameter_dict["vary"].append(parameter.vary)
-            parameter_dict["non-negative"].append(parameter.non_negative)
-            parameter_dict["expression"].append(parameter.expression)
-        return pd.DataFrame(parameter_dict)
+        return pd.DataFrame(self.to_parameter_dict_list(as_optimized=as_optimized))
+
+    def get_group_for_paramter_by_label(
```
joernweissenborn, line 265:
```suggestion
    def get_group_for_parameter_by_label(
```

glotaran/parameter/parameter.py
```diff
@@ -409,6 +462,50 @@ def set_value_from_optimization(self, value: float):
         """
         self.value = np.exp(value) if self.non_negative else value
 
+    def markdown(
```
s-weigand, line 465: This will be refactored in a later PR and properly tested. (Ref.: call with @joernweissenborn )

glotaran/parameter/test/test_parameter.py
```diff
@@ -488,97 +346,26 @@ def test_parameter_numpy_operations():
...
-    b:
-        - ["1", 0.25, {vary: false, min: 0, max: 8}]
-        - ["2", 0.75, {expr: '1 - $b.1', non-negative: true}]
-    rates:
-        - ["total", 2]
-        - ["branch1", {expr: '$rates.total * $b.1'}]
-        - ["branch2", {expr: '$rates.total * $b.2'}]
-    """,
-        format_name="yml_str",
+def test_parameter_to_from_dict():
```
s-weigand, line 349:
```suggestion
def test_parameter_dict_roundtrip():
```
This might be a more meaningful name.

s-weigand (approved):
LGTM just some leftover nitpicks 👍

glotaran/parameter/parameter.py
```diff
@@ -42,11 +43,12 @@ def __init__(
         self,
         label: str = None,
         full_label: str = None,
-        expression: str = None,
-        maximum: int | float = np.inf,
-        minimum: int | float = -np.inf,
+        expression: str | None = None,
```
jsnel, line 46: Seems label and full_label should have the same type annotation here.

glotaran/parameter/parameter.py
```diff
@@ -144,6 +148,55 @@ def from_list_or_value(
             param._set_options_from_dict(options)
         return param
 
+    @classmethod
+    def from_dict(cls, parameter_dict: dict[str, Parameter]) -> Parameter:
```
jsnel, line 152: This method needs to be revised, very confusing what the intention is here.
The input parameter_dict should not be manipulated internally with the same name.

glotaran/parameter/parameter.py
```diff
@@ -144,6 +148,55 @@ def from_list_or_value(
...
+        -------
+        Parameter
+            The created :class:`Parameter`
+        """
+        parameter_dict = {k.replace("-", "_"): v for k, v in parameter_dict.items()}
+        parameter_dict["full_label"] = parameter_dict["label"]
+        parameter_dict["label"] = parameter_dict["label"].split(".")[-1]
+        return cls(**parameter_dict)
+
+    def as_dict(self, as_optimized: bool = True) -> dict:
```
jsnel, line 172: Wrong typing. Also check consistency by doing as_dict(from_dict(x))

glotaran/parameter/parameter_group.py
```diff
@@ -69,7 +70,7 @@ def __init__(self, label: str = None, root_group: ParameterGroup = None):
     @classmethod
     def from_dict(
         cls,
-        parameter_dict: dict[str, dict | list],
+        parameter_dict: dict[str, dict[str, Any] | list[float | list[Any]]],
```
jsnel, line 73: How to type this properly.

glotaran/parameter/parameter.py
s-weigand: "Fixed" in 053e356

glotaran/parameter/parameter.py
s-weigand: The typing is "fixed" in 053e356
As for overwriting, since only the keys get changed this would be a case where reassigning a value is ok.
Typing wise we should most likely use some kind of `Mapping` + `Protocol` mix, but I think fixing our typing issues on a fundamental level is a bit out of scope for this PR.

glotaran/parameter/parameter_group.py
s-weigand: This is something we should do in our new `glotaran.typing` package and will need thinking due to the nested structure of `ParameterGroup` and all the different ways to represent a `Parameter`. Most likely some kind of `Mapping` + `Protocol` mix.

glotaran/parameter/parameter.py
joernweissenborn: Because they are `str`? I don't understand the issue...

jsnel (approved):
Reviewed ok.

glotaran/parameter/parameter.py
s-weigand: It is also possible that they are `None` in case of `full_label` it will be set by the setter to `str(None)` which is `'None'` so this is actually right. But `label` can be `str | None` all the time.

jsnel: It's just a minor (nitpicking) thing, but I mean if you make a default values None, the type hint should also include that type, just like with 
`expression: str | None = None,`
it should be
```
label: str | None = None,
full_label: str | None = None,
```

s-weigand (approved):
LGTM 👍

PR #912, opened 2021-11-25
🩹 Remove inheriting from 'FileLoadableProtocol'

s-weigand:
In python 3.9.7 (currently the only python 3.9 version on conda) [there was a bug](https://github.com/python/cpython/pull/27545) that ignores the __init__ method of classes that inherit from a Protocol.

This lead to the following error (discovered by @jsnel )
```python
TypeError: Protocols cannot be instantiated
```

Rather than restricting the usable python versions for typing sake, I will file an issue at `mypy` since the conformity with protocols should be picked up automatically.

### Change summary

- 🩹 Removed inheriting from 'FileLoadableProtocol'

### Checklist


joernweissenborn (approved):
lgtm

jsnel (approved):
lgtm

PR #914, opened 2021-11-26
Refactored ModelProperty.

### LINKED ISSUE #901: 🐛 Parameters used for expressions missing in Model markdown and report
### Version information

- pyglotaran version: All
- Python version: All
- Operating System: All

### Describe the bug

In the markdown report, only Parameters directly used in the model are displayed, but if users define parameters only for expressions those are missing making the report incomplete.
In addition, the parameters used only show their numerical value and not the expression used to calculate them.

### To Reproduce

You can edit the quickstart to use 

```yaml
kinetic: 
    - 0.5
    - 0.3
    - ["3", {expr: "$kinetic.expression_parameter"}]
    - ["expression_parameter", 0.1]

```

Instead of

https://github.com/glotaran/pyglotaran/blob/ddec540e506ff6e80c49c09aa53e3ad4ef6e3e46/docs/source/notebooks/quickstart/parameters.yaml#L5-L9

For quick confirmation this can be done [on binder](https://mybinder.org/v2/gh/glotaran/pyglotaran.git/main?urlpath=lab%2Ftree%2Fdocs%2Fsource%2Fnotebooks)

### Expected behavior

Show expressions used to calculate parameters if they use expressions and all parameters needed to evaluate the expression.

### Screenshots
The example mentioned above run in binder.
grafik
grafik

joernweissenborn:
This PR refactors the `ModelProperty` class.

### Change summary

- Markdown printing of `Model` has been changed
  - Beautification is left for followup PR

### Checklist

### Closes issues

closes #901

### DIFF
--- a/glotaran/model/property.py
+++ b/glotaran/model/property.py
@@ -1,8 +1,9 @@
-"""The model property class."""
+"""This module holds the model property class."""
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import Callable
 from typing import Mapping
 from typing import Sequence
 from typing import TypeVar
@@ -24,10 +25,31 @@ ParameterOrLabel = TypeVar("ParameterOrLabel", str, Parameter)
 
 
 class ModelProperty(property):
+    """ModelProperty is an extension of the property decorator.
+
+    It adds convenience functions for meta programming model items.
+    """
+
     def __init__(
         self, cls: type, name: str, property_type: type, doc: str, default: Any, allow_none: bool
     ):
+        """Create a new model property.
 
+        Parameters
+        ----------
+        cls : type
+            The class the property is being attached to.
+        name : str
+            The name of the property.
+        property_type : type
+            The type of the property.
+        doc : str
+            A documentation string of for the property.
+        default : Any
+            The default value of the property.
+        allow_none : bool
+            Whether the property is allowed to be None.
+        """
         self._name = name
         self._allow_none = allow_none
         self._default = default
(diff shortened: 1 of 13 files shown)


glotaran/model/test/test_model_property.py
```diff
@@ -0,0 +1,133 @@
+"""Tests for glotaran.model.dataset_model.DatasetModel"""
```
joernweissenborn, line 1:
```suggestion
"""Tests for glotaran.model.property.ModelProperty"""
```

glotaran/model/property.py
```diff
@@ -3,85 +3,168 @@
...
+
+def _subtype(t: type) -> type:
+    if _is_sequence_type(t):
+        return t.__args__[0]
+    elif _is_mapping_type(t):
+        return t.__args__[1]
+    return t
+
+
+def glotaran_is_parameter_property(self) -> bool:
```
joernweissenborn, line 47: Note to myself: remove this

glotaran/model/property.py
```diff
@@ -3,85 +3,168 @@
...
+        return value
+
+    return getter
+
+
+def _model_property_setter_factory(cls: type, model_property: ModelProperty):
+    @wrap_func_as_method(cls, name=model_property._name)
+    def setter(self, value: model_property.glotaran_property_type):
+        if value is None and not model_property._allow_none:
+            raise Exception(
```
joernweissenborn, line 66: Note to myself: use value error

glotaran/parameter/parameter.py
```diff
@@ -485,22 +485,23 @@ def markdown(
...
-        if self.vary:
-            if self.standard_error is not np.nan:
-                value += f"±{self.standard_error}"
-            if initial_parameter is not None:
-                initial_value = initial_parameter.get(self.full_label).value
+        parameter = self if all_parameters is None else all_parameters.get(self.full_label)
+        value = f"{parameter.value:.2e}"
+        if parameter.vary:
+            if parameter.standard_error is not np.nan:
+                value += f"±{parameter.standard_error}"
```
s-weigand, line 492:
```suggestion
                value += f"±{parameter.standard_error:.2e}"
```
Currently the standard error is for example rendered like this:
`('s2', 's1'): kinetic.1(5.00e-01±7.284630572763264e-05, initial: 5.00e-01)`

The other option would be to make the significant numbers an option with a sensible default.

s-weigand (approved):
Functionality wise this looks good to me 👍 
So thanks a lot, users will love the enhanced information they get. ❤️ 

**But** also it works quite well for `0.6.0` we should really reflect on if we really need this kind of metaprogramming or if we could maybe solve our requirements in a more maintainable fashion and pythonic way, instead of our mad scientist (e.g. utilizing [attrs](https://www.attrs.org/en/stable/)).

For me the most problematic point is inheriting a from the `property` decorator class, which leads this strange tests in `test_model_property` that don't reflect the actual usage.
The more tests the married, but they should reflect the usage in the code.

For now, there is no quick way around it (at least after 2h of trial and error from my side) and it would be out of scope for this PR, since this is an integral part of the code for 3 years.

s-weigand (approved):

jsnel (approved):
LGTM

PR #930, opened 2021-12-03
Add include statement to model yml

joernweissenborn:
This PR adds the statement `include` to the model spec. This is useful for splitting large models into multiple files.

### Change summary

- added `include` keyword to model spec yaml and added parsing logic
- this might have troubles when round tripping, so more work on that is required

### Checklist

### Closes issues

closes #XXXX

### DIFF
--- a/glotaran/builtin/io/yml/test/test_model_include.py
+++ b/glotaran/builtin/io/yml/test/test_model_include.py
@@ -1,0 +1,40 @@
+from pathlib import Path
+
+from glotaran.io import load_model
+
+TEST_MODEL_DATASET = """\
+dataset:
+  d1:
+    megacomplex: [m1]
+"""
+TEST_MODEL_MEGACOMPLEX = """\
+megacomplex:
+  m1:
+    type: decay
+    k_matrix: []
+"""
+
+TEST_MODEL = """\
+include:
+  - {d_path}
+  - {m_path}
+"""
+
+
+def test_model_include(tmp_path: Path):
+    d_path = tmp_path / "dataset.yml"
+    m_path = tmp_path / "megacomplex.yml"
+    model_path = tmp_path / "model.yml"
+
+    with open(d_path, "w") as f:
+        f.write(TEST_MODEL_DATASET)
+
+    with open(m_path, "w") as f:
+        f.write(TEST_MODEL_MEGACOMPLEX)
+    with open(model_path, "w") as f:
+        f.write(TEST_MODEL.format(d_path=d_path, m_path=m_path))
+
+    model = load_model(model_path)
+
+    assert "d1" in model.dataset
+    assert "m1" in model.megacomplex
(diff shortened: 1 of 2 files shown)


jsnel:
@s-weigand please review and consider how this may interact with the roundtrip support recently introduced by switching to ruaml for yaml parsing: https://github.com/glotaran/pyglotaran/pull/893

glotaran/builtin/io/yml/test/test_model_include.py
```diff
@@ -0,0 +1,40 @@
...
+    m_path = tmp_path / "megacomplex.yml"
+    model_path = tmp_path / "model.yml"
+
+    with open(d_path, "w") as f:
+        f.write(TEST_MODEL_DATASET)
+
+    with open(m_path, "w") as f:
+        f.write(TEST_MODEL_MEGACOMPLEX)
+    with open(model_path, "w") as f:
+        f.write(TEST_MODEL.format(d_path=d_path, m_path=m_path))
```
s-weigand, line 35: Writing to file could be done by using `Path.write_text` (which internally also uses a context manager).

glotaran/builtin/io/yml/yml.py
```diff
@@ -48,6 +48,13 @@ def load_model(self, file_name: str) -> Model:
 
         spec = sanitize_yaml(spec)
 
+        if "include" in spec:
+            includes = spec.pop("include")
+            if not isinstance(includes, list):
+                includes = [includes]
+            for include in includes:
+                spec.update(self._load_yml(include))
```
s-weigand, line 56: We should check if there are duplicated entries and at least warn users if they are going to be overwritten.
Atm, we just overwrite things by order, and even more dangerous we overwrite at the top level! 
Image the following scenario:
A user defines each megacomplex in its own file and includes both, all but the last one get overwritten without any warning.

```python
In [1]: spec = {"megacomplex": {"m1": {"foo":1}}}

In [2]: mc_include = {"megacomplex": {"m2": {"bar":2}}}

In [3]: spec.update(mc_include)

In [4]: spec
Out[4]: {'megacomplex': {'m2': {'bar': 2}}}
```

s-weigand (changes requested):
I know this is just the first quick and dirty implementation, but I think the way it is atm it will cause more problems than it solves.
In general, a feature like this is very useful to split up logic, one could even imagine a kind of database with known megacomplexes users can just use to build their model.

Using a YAML tag (`!include`) would give us more flexibility, e.g. see this dated answer from the `ruaml` Author
https://stackoverflow.com/a/44913652/3990615

In any case we need to better handle users accidentally overwriting parts of their specs!

As for round trip support, this will need some investigation and extra work in any case since the ruaml updating doesn't support direct updating from dict w/o loosing all comments.

s-weigand:
Closing in favor #955

PR #931, opened 2021-12-03
🩹 Fix save_result crashes on Windows if input data are on a different drive than result

### LINKED ISSUE #924: 🐛 save_result crashes on Windows if input data are on a different drive than result
### Version information

- pyglotaran version: 0.5.0
- Python version: all
- Operating System: Windows

### Describe the bug

While upgrading the `pyglotaran-examples` not to use deprecated functionality, I encountered a serious bug where `save_result` crashes when saving the scheme.

```python
ValueError: path is on mount 'D:', start on mount 'C:'
```

The problem is that I have the `pyglotaran-examples` at `D:\git` and the results are saved under `%USERPROFILE%` which is on the `C:\` drive which lets `relpath` crash.

https://github.com/glotaran/pyglotaran/blob/89046e0a95117d63336f5cdb637bf8acf52addb2/glotaran/project/dataclass_helpers.py#L197-L201

https://github.com/glotaran/pyglotaran/blob/89046e0a95117d63336f5cdb637bf8acf52addb2/glotaran/utils/io.py#L181

### To Reproduce

Copy the `pyglotaran-examples` to a different drive than the `C:\` drive and run:

```console
python scripts/run_examples.py sim-3d-disp 
```

### Expected behavior

No crash!

### Additional context

We have multiple options to fix this:
1. Copy the data to a path relative to `scheme.yml` and warn the user
2. Use absolute paths and warn the user
3. Overwrite `scheme.data` with `result.data`, since the result data also contain the original data and are saved relative to `scheme.yml`

@joernweissenborn @jsnel  What do you think?

`1.` or `2.` (or another solution) need to be implemented to make `file_loadable_field` robust, `3.` is optional and a disk space-saving option if we choose to go with `1.`.

s-weigand:
This fixes the crash of `save_result` when the original data and the result path are on different drives on windows.

~The commit e431760a9646e3ca311107926cafa4a952092416 is optional.~

### Change summary

- 🩹 Made 'relative_posix_path' more stable on windows
~- ♻️ Changed save_result folder to use the result.data for scheme.data~

### Checklist

### Closes issues

closes #924

### DIFF
--- a/glotaran/utils/io.py
+++ b/glotaran/utils/io.py
@@ -184,5 +184,5 @@ def relative_posix_path(source_path: StrOrPath, base_path: StrOrPath | None = No
         try:
             source_path = os.path.relpath(source_path, Path(base_path).as_posix())
         except ValueError:
-            source_path = Path(source_path).as_posix()
+            pass
     return Path(source_path).as_posix()
--- a/glotaran/utils/io.py
+++ b/glotaran/utils/io.py
@@ -164,6 +164,9 @@ def load_datasets(dataset_mappable: DatasetMappable) -> DatasetMapping:
 def relative_posix_path(source_path: StrOrPath, base_path: StrOrPath | None = None) -> str:
     """Ensure that ``source_path`` is a posix path, relative to ``base_path`` if defined.
 
+    On windows if ``source_path`` and ``base_path`` are on different drives, it will return
+    the absolute posix path to the file.
+
     Parameters
     ----------
     source_path : StrOrPath
@@ -178,5 +181,8 @@ def relative_posix_path(source_path: StrOrPath, base_path: StrOrPath | None = No
     """
     source_path = Path(source_path).as_posix()
     if base_path is not None and os.path.isabs(source_path):
-        source_path = os.path.relpath(source_path, Path(base_path).as_posix())
+        try:
+            source_path = os.path.relpath(source_path, Path(base_path).as_posix())
+        except ValueError:
+            source_path = Path(source_path).as_posix()
     return Path(source_path).as_posix()
--- a/glotaran/utils/io.py
+++ b/glotaran/utils/io.py
@@ -164,7 +164,7 @@ def load_datasets(dataset_mappable: DatasetMappable) -> DatasetMapping:
 def relative_posix_path(source_path: StrOrPath, base_path: StrOrPath | None = None) -> str:
     """Ensure that ``source_path`` is a posix path, relative to ``base_path`` if defined.
 
-    On windows if ``source_path`` and ``base_path`` are on different drives, it will return
+    On Windows if ``source_path`` and ``base_path`` are on different drives, it will return
     the absolute posix path to the file.
 
     Parameters
--- a/glotaran/utils/test/test_io.py
+++ b/glotaran/utils/test/test_io.py
@@ -160,9 +160,9 @@ def test_relative_posix_path(tmp_path: Path, rel_file_path: str):
     assert rel_result_no_coomon == f"../{rel_file_path}"
 
 
-@pytest.mark.skipif(not sys.platform.startswith("win32"), reason="Only needed for wondows")
+@pytest.mark.skipif(not sys.platform.startswith("win32"), reason="Only needed for Windows")
 def test_relative_posix_path_windows_diff_drives():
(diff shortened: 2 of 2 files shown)


joernweissenborn:
Using `result.data` will crash in future.

s-weigand:
I dropped e431760 since [it causes the examples to fail](https://github.com/glotaran/pyglotaran/runs/4413109109?check_suite_focus=true) when reoptimizing a result for some reason.
Since the result contains the original data, this needs some more investigation!

glotaran/utils/test/test_io.py
```diff
@@ -157,3 +158,13 @@ def test_relative_posix_path(tmp_path: Path, rel_file_path: str):
     )
 
     assert rel_result_no_coomon == f"../{rel_file_path}"
+
+
+@pytest.mark.skipif(not sys.platform.startswith("win32"), reason="Only needed for wondows")
```
jsnel, line 163: I understand you hold a grudge, but I still suggest calling Windows by it's correct name. ^^

glotaran/utils/test/test_io.py
```diff
@@ -157,3 +158,13 @@ def test_relative_posix_path(tmp_path: Path, rel_file_path: str):
...
+def test_relative_posix_path_windows_diff_drives():
+    """os.path.relpath doesn't cause crash when files are of different drives."""
```
jsnel, line 165: when files are *on* different drives

glotaran/utils/test/test_io.py
jsnel: Perhaps do a quick investigation on the behavior on platforms other than Windows, e.g. a Mac user that has their data on an external drive or in the cloud.

glotaran/utils/io.py
```diff
@@ -164,6 +164,9 @@ def load_datasets(dataset_mappable: DatasetMappable) -> DatasetMapping:
 def relative_posix_path(source_path: StrOrPath, base_path: StrOrPath | None = None) -> str:
     """Ensure that ``source_path`` is a posix path, relative to ``base_path`` if defined.
 
+    On windows if ``source_path`` and ``base_path`` are on different drives, it will return
```
jsnel, line 167: windows -> Windows

glotaran/utils/test/test_io.py
```diff
@@ -1,6 +1,7 @@
 """Tests for glotaran/utils/io.py"""
 from __future__ import annotations
 
+import sys
```
jsnel, line 4: is there perhaps a bigger role for [pathlib](https://docs.python.org/3/library/pathlib.html) in lieu of sys?

jsnel (changes requested):
Some typos to correct and suggest to do some sanity investigation on the other platforms we target.

glotaran/utils/test/test_io.py
s-weigand: Ups will fix that typo.

As per [documentation](https://docs.python.org/3/library/os.path.html#os.path.relpath), this is a windows only thing.

glotaran/utils/test/test_io.py
s-weigand: Fixed in 4d6301c0a66e31de369eba0c044c162f2f81c60e

glotaran/utils/test/test_io.py
s-weigand: Fixed in 4d6301c0a66e31de369eba0c044c162f2f81c60e

glotaran/utils/test/test_io.py
s-weigand: Fixed in 4d6301c0a66e31de369eba0c044c162f2f81c60e

glotaran/utils/io.py
s-weigand: Fixed in 4d6301c0a66e31de369eba0c044c162f2f81c60e

glotaran/utils/test/test_io.py
s-weigand: Could you elaborate more on what you mean?

jsnel: In this case sys is only used for host os resolving, so it's fine.

jsnel (approved):

PR #948, opened 2022-01-04
🩹Fix bug in create_index_independent_result_dataset

### LINKED ISSUE #950: 🐛 Crash in optimization_group_calculator_linked when using guidance spectra
### Version information

- pyglotaran version: main

### Describe the bug

The code
```python
   dataset["matrix"] = (
                (
                    (self._model_dimension),
                    ("clp_label"),
                ),
                self._group.matrices[label].matrix,
            )
```
doesn't work for an analysis with multiple dataset, including some guidance spectra (it crashes on creating the result for the first guidance spectra) because self._group.matrices then returns a list of CalculatedMatrix objects, instead of a single CalculatedMatrix object.

A fix would be

```python
 dataset["matrix"] = (
                (
                    (self._global_dimension),
                    (self._model_dimension),
                    ("clp_label"),
                ),
                np.asarray([m.matrix for m in self._group.matrices[label]]),
            )
```
but only in the case where the result is a list of CalcualtedMatrix objects.

### To Reproduce

Run the Artificial Lego example.

### Expected behavior

No crashes, while also not failing for all existing examples and unit tests.

### Possible fix

Testing what the type is and excuting case dependent code.

```python
 dummy = self._group.matrices[label]
        if isinstance(dummy, CalculatedMatrix):
            dataset["matrix"] = (
                (
                    (self._model_dimension),
                    ("clp_label"),
                ),
                self._group.matrices[label].matrix,
            )
        else:
            dataset["matrix"] = (
                (
                    (self._global_dimension),
                    (self._model_dimension),
                    ("clp_label"),
                ),
                np.asarray([m.matrix for m in self._group.matrices[label]]),
            )
```

jsnel:
### Change summary

- Indexing did not match content of _group.matrices (list of CalculatedMatrix objects)
- Dimension did not match (for guidance spectra) (global_dimension_index appeared to be missing

### Checklist

### Closes issues

Closes #950

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,8 @@
 
 ### 🩹 Bug fixes
 
+- 🩹 Fix Crash in optimization_group_calculator_linked when using guidance spectra (#950)
+
 ### 📚 Documentation
 
 ### 🗑️ Deprecations (due in 0.8.0)
--- a/glotaran/analysis/optimization_group_calculator_linked.py
+++ b/glotaran/analysis/optimization_group_calculator_linked.py
@@ -463,10 +463,11 @@ class OptimizationGroupCalculatorLinked(OptimizationGroupCalculator):
 
         dataset["matrix"] = (
             (
+                (self._global_dimension),
                 (self._model_dimension),
                 ("clp_label"),
             ),
-            self._group.matrices[label].matrix,
+            np.asarray([m.matrix for m in self._group.matrices[label]]),
         )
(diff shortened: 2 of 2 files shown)


jsnel:
In lieu of review by another code owner (@joernweissenborn), considering all unit tests are passing I will use my administrator privileges to merge in the patch so @ism200 can proceed with his case study. If a more elegant solution is available that can be addressed in a future PR.

PR #958, opened 2022-01-16
🚧📚 Add changes of PR 860 to changelog

s-weigand:
Just adding #860 to the changelog which we forgot in the actual PR.

### Change summary

- 🚧📚 Added changes of PR 860 to changelog

### Checklist

- [x] ✔️ Passing the tests (mandatory for all PR's)
- [x] 🚧 Added changes to changelog (mandatory for all PR's)

PR #959, opened 2022-01-16
🔧🚇 Exclude test files from duplication checks on sonarcloud

s-weigand:
Due to duplications that aren't even part of a PR sonarcloud is failing its test (e.g. #958 )

image

image

By adding the test files which should be verbose and readable, I hope we can get around this issue.

### Change summary

- 🔧 Exclude test files from duplication checks

### Checklist

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,1 @@
+sonar.cpd.exclusions=**/test_*.py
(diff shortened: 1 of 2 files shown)


s-weigand:
Much better ^^
image

s-weigand:
📚 As future reference where to find the config options, [this is the page in their docks](https://docs.sonarcloud.io/advanced-setup/analysis-scope)

jsnel (approved):
Good idea.  👍

PR #966, opened 2022-01-24
👌 Add SavingOptions to save_result API 

s-weigand:
My take on #874 😝 

The main difference is `data_filter` is applied on the `save_result` level instead of adding an unused keyword to all `save_dataset` methods.
I also deprecated the `legacy` plugin and made `folder` throw a warning when not used from inside another plugin, thus promoting `yml` to the goto `save_result` plugin, which will make the transition to project easier.
The `folder` plugin now also doesn't save `model.yml` nor `scheme.yml`, that way it can easier be reused in different plugins.

### Change summary

-  🚧 Salvage parts of PR 874 
- 🧪 Removed part of a test which fails and should fail 
- 👌 Restored wanted functionality
- 👌 Homogenized save_result return value to always be a list of strings
- 👌 Properly reexport functionality from glotaran.deprecation
- 🗑️ Deprecated save_result legacy plugin and made folder plugin internal use only

### Checklist

### Closes issues

closes #874

### DIFF
--- a/glotaran/builtin/io/folder/folder_plugin.py
+++ b/glotaran/builtin/io/folder/folder_plugin.py
@@ -8,11 +8,12 @@ from __future__ import annotations
 
 from pathlib import Path
 from typing import TYPE_CHECKING
+from warnings import warn
 
+from glotaran.deprecation import warn_deprecated
 from glotaran.io import save_dataset
-from glotaran.io import save_model
 from glotaran.io import save_parameters
-from glotaran.io import save_scheme
+from glotaran.io import save_result
 from glotaran.io.interface import ProjectIoInterface
 from glotaran.plugin_system.project_io_registration import SAVING_OPTIONS_DEFAULT
 from glotaran.plugin_system.project_io_registration import register_project_io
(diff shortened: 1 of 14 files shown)


joernweissenborn (approved):
lgtm

glotaran/builtin/io/folder/folder_plugin.py
```diff
@@ -74,6 +137,15 @@ def save_result(
         ValueError
             If ``result_path`` is a file.
         """
+        if used_inside_of_plugin is not True:
+            warn(
+                UserWarning(
+                    "The folder plugin is only intended for internal use by other plugins "
+                    "as quick way to save most of the files. The saved result will be incomplete "
+                    ", thus it is not recommended to be used directly."
```
jsnel, line 145: In general, better move punctuation like `,` to end of a line, now it's hard to see that you have an extra space 
`UserWarning: The folder plugin is only intended for internal use by other plugins as quick way to save most of the files. The saved result will be incomplete , thus it is not recommended to be used directly.`

jsnel (dismissed):
Reviewed ok.

glotaran/builtin/io/folder/folder_plugin.py
s-weigand: Fixed in 72ca809

PR #967, opened 2022-01-24
🔧🚇 Only run check-maifest on the CI

s-weigand:
While it is important that we check if all files we want to bundle are actually bundled in the distribution, checking this in pre-commit can be cumbersome.
E.g. you were working on a PR and when you got it to work the way you want you to commit different small changes in contextual closed commits. If you added new files which you don't want to add in this commit `check-manifest` won't allow you to make this commit since it looks at all files and not only the tracked ones.
This in turn means that you need to make big commits or use the `--no-verify` flag which both can lead to bad commits.

### Change summary

- 🔧🚇 Only run check-manifest on the CI

### Checklist

### DIFF
--- a/.github/workflows/CI_CD_actions.yml
+++ b/.github/workflows/CI_CD_actions.yml
@@ -13,6 +13,21 @@ jobs:
       - name: Run pre-commit
         uses: pre-commit/action@v2.0.3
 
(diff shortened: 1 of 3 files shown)


jsnel (approved):
Good idea. 👍

PR #977, opened 2022-01-29
✨Add Python 3.10 support

s-weigand:
[Numba 0.55.0 added support for python 3.10](https://github.com/numba/numba/releases/tag/0.55.0) which means so can we

### Change summary

- 🚇🧪 Add tests for python 3.10
- ⬆️ Raised max supported python version to <3.11
- ⬆️ Unpin `setuptools`

Using an old version of `setuptools` caused a lot of import errors in python 3.10
Pinning the version of `setuptools` goes back to the time when we still had c-extensions, so it is safe to unpin it

### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,27 +1,23 @@
 version: 2
 updates:
-- package-ecosystem: pip
-  directory: "/"
-  schedule:
-    interval: weekly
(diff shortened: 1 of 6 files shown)


jsnel (approved):
LGTM. 

It's about time. 😅

PR #979, opened 2022-01-29
🚇 Only run doxygen workflow when requested

s-weigand:
Since we most of the time don't make use of the Doxygen generated callgraphs we don't need to always run the workflow.
Thus is created a [github environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) called `Debug` which needs approval from a member of @glotaran/contributors to run.
That way Doxygen won't run until we approve/request it.

### Change summary

- 🚇 Only run doxygen workflow when requested

### Checklist


jsnel (approved):
👍

PR #981, opened 2022-01-30
🚇👌 Run Doxygen workflow only on  'workflow_dispatch'

s-weigand:
As requested by @jsnel this changes the Doxygen workflow to truly only run when requested.
The problem with the current implementation (#979 ) via github environment is that the UI never gets to the finished green checkmark state, but waits forever for the workflow to be approved.

grafik

This makes it hard to spot which PR is ready to be merged in the PR overview.

grafik

The new implementation allows specifying a repository to build the oxygen docs for (default `glotaran/pyglotaran` but could also be a fork) and a pr nr (default "" which uses the branch used to dispatch the workflow from).

### Change summary

- 🚇👌 Reworked doxygen workflow to be 'workflow_dispatch' but work with PRs

### Checklist

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -1,18 +1,62 @@
 name: "Doxygen"
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  workflow_dispatch:
+    inputs:
+      pr_nr:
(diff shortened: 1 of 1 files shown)


jsnel (approved):
,👌

PR #983, opened 2022-01-30
✨Make guidance spectra a first class citizen

### LINKED ISSUE #846: ✨ Create Clp Guidance
### Is your feature request related to a problem? Please describe.

This is a request by @ism200. Basically, sometimes a virtual dataset is added to the analysis to provide guidance for the clp estimation.

### Describe the solution you'd like

This is tricky to solve. Semantically, the guidance spectrum is dataset by it's own. This means it has to be representented as a special dataset.

An idea woud be to introduce an option for a dataset model called `clp_guide`. This will mean that this dataset is allowed to have no megacomplexes (ectually forbitten to have one). It will just return an array of ones, depending of it is index dependent it is one.

I think this is best solved along the path to make mixed nnls and varpro fitting.

### Additional context

There should be a case study for that.

s-weigand:
This is just a preview for maintainers

### Change summary

- ✨ Initial draft for a simplified specification of guidance spectra
- ♻️🩹 Renamed overwrite_{model|global}_dimension to force_{model|global}_dimension
- 🩹 Fixed tests
- ♻️ Renamed attributes force_{model|global}_dimension back to overwrite_{model|global}_dimension
- ♻️ Renamed methods overwrite_{model|global}_dimension to replace_{model|global}_dimension

### Checklist

### Closes issues

closes #846


joernweissenborn (approved):
lgtrm

glotaran/model/dataset_model.pyi
```diff
@@ -16,6 +16,10 @@ def create_dataset_model_type(properties: dict[str, Any]) -> type[DatasetModel]:
 class DatasetModel:
 
     label: str
+    clp_guidance: str | None
+    overwrite_model_dimension: str | None
+    overwrite_global_dimension: str | None
```
jsnel, line 21: Change overwrite to override.

glotaran/model/dataset_model.py
```diff
@@ -41,9 +42,13 @@ def iterate_megacomplexes(
         self,
     ) -> Generator[tuple[Parameter | int | None, Megacomplex | str], None, None]:
         """Iterates of der dataset model's megacomplexes."""
```
jsnel, line 44: of der -> over the
-> dataset_models's megacomplexes

glotaran/model/model.py
```diff
@@ -34,12 +34,15 @@
 }
 
 default_dataset_properties = {
+    "clp_guidance": {"type": str, "allow_none": True},
+    "overwrite_model_dimension": {"type": str, "allow_none": True},
+    "overwrite_global_dimension": {"type": str, "allow_none": True},
```
jsnel, line 39: override

jsnel (changes requested):
Naming things ...

glotaran/model/dataset_model.pyi
```diff
@@ -30,9 +34,9 @@ class DatasetModel:
     ) -> Generator[tuple[Parameter | int | None, Megacomplex | str], None, None]: ...
     def get_model_dimension(self) -> str: ...
     def finalize_data(self, dataset: xr.Dataset) -> None: ...
-    def overwrite_model_dimension(self, model_dimension: str) -> None: ...
+    def replace_model_dimension(self, model_dimension: str) -> None: ...
```
joernweissenborn, line 37: `set_model_dimension` would fit better

s-weigand: Fixed in [57c32a9](https://github.com/glotaran/pyglotaran/pull/983/commits/57c32a9b7c12ea41a18862f8c621615b3dcc0a80)

glotaran/model/model.py
```diff
@@ -34,12 +34,15 @@
 }
 
 default_dataset_properties = {
+    "clp_guidance": {"type": str, "allow_none": True},
```
joernweissenborn, line 37: A little nitpicky: I personally like to have an explicit boolean switch like `is_clp_guide` instead of implicit enabling by setting this to not `None`. there might be more options for guidance in future and then implicit eanabling becomes a problem. so I think there should be a more explicit way to enable it.

Another idea, which need discussion: we could intruce a `type` property to datasetmodels, which defaults to `dataset`, but could be clp guide or e.g. a measured irf, But is for another PR, since that ois the whole problem of how to add extra data to your analysis.

s-weigand: IMHO adding `is_clp_guide` would be superfluous and make the API more boilerplate cluttered, which this PR tries to reduce. Rather than cluttering the API, I would inform the user that they did something wrong if they do.

@jsnel what are your thoughts on this?

glotaran/model/dataset_model.py
s-weigand: Fixed in [4c836d6](https://github.com/glotaran/pyglotaran/pull/983/commits/4c836d6b642e40f656da3ba823ffe6bf6625f453)

glotaran/model/model.py
s-weigand: Fixed in [7f565aa](https://github.com/glotaran/pyglotaran/pull/983/commits/7f565aac69c6433bfdc04444c3b2f0dd1997b31f)

glotaran/model/dataset_model.pyi
s-weigand: Fixed in [7f565aa](https://github.com/glotaran/pyglotaran/pull/983/commits/7f565aac69c6433bfdc04444c3b2f0dd1997b31f)

glotaran/model/model.py
jsnel: The addition of an explicit boolean switch like is_clp_guide doesn't make sense to me, but I like the idea of introducing a type to dataset models at some point. However, I also agree this is best introduced in a subsequent PR where that type is actually needed. 

For now I think the current form of introduction is fine as-is. If needed it can later be easily deprecated in favor of the type argument. So instead of:
```yml
clp_guidance: s1
```
you would get
```yml
type: clp_guidance
compartments: s1
```

s-weigand: To make it more generic we could even use
```yaml
type: clp_guidance
clp_label: s1
```

s-weigand:
As discussed with @joernweissenborn today he will implement this as a builtin megacomplex because it will make it more consistent with the modeling language spec and the theory description in "the book" (his thesis).

s-weigand:
Superseded by #1029

PR #994, opened 2022-02-06
Added irf shift to result dataset.

### LINKED ISSUE #987: ✨ Store Irf Shift in Result
This is needed for debugging. since one has a long list of parameters for this, it is crucial to plot this to evaluate the result of an optimization.

Implementation should be be done be by implementing a `finalize_data' method for irfs which all megacomplexes using irfs must call on `finalize_data`. A problem is that when here are multiple megacomplexes using the same irf, so this must be handled like SAS.

joernweissenborn:
The irfs shifted centers are stored in the result now.

### Checklist

### Closes issues

closes #987

### DIFF
--- a/glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
+++ b/glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
@@ -7,6 +7,7 @@ import xarray as xr
 
 from glotaran.builtin.megacomplexes.decay.irf import Irf
 from glotaran.builtin.megacomplexes.decay.irf import IrfMultiGaussian
+from glotaran.builtin.megacomplexes.decay.util import retrieve_irf
 from glotaran.model import DatasetModel
 from glotaran.model import Megacomplex
 from glotaran.model import ModelError
@@ -84,6 +85,7 @@ class CoherentArtifactMegacomplex(Megacomplex):
                 (global_dimension, "coherent_artifact_order"),
                 dataset.clp.sel(clp_label=self.compartments()).values,
             )
+        retrieve_irf(dataset_model, dataset, dataset_model.get_global_dimension())
 
 
 @nb.jit(nopython=True, parallel=True)
--- a/glotaran/builtin/megacomplexes/decay/util.py
+++ b/glotaran/builtin/megacomplexes/decay/util.py
@@ -103,8 +103,7 @@ def finalize_data(
             is_full_model,
             as_global,
         )
-    if isinstance(dataset_model.irf, IrfMultiGaussian) and "irf" not in dataset:
-        retrieve_irf(dataset_model, dataset, global_dimension)
+    retrieve_irf(dataset_model, dataset, global_dimension)
 
     if not is_full_model:
         multiple_complexes = len(decay_megacomplexes) > 1
@@ -314,6 +313,8 @@ def retrieve_decay_associated_data(
 
 
 def retrieve_irf(dataset_model: DatasetModel, dataset: xr.Dataset, global_dimension: str):
+    if not isinstance(dataset_model.irf, IrfMultiGaussian) or "irf" in dataset:
+        return
 
     irf = dataset_model.irf
     model_dimension = dataset_model.get_model_dimension()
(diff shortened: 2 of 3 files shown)


glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
```diff
@@ -84,6 +85,7 @@ def finalize_data(
                 (global_dimension, "coherent_artifact_order"),
                 dataset.clp.sel(clp_label=self.compartments()).values,
             )
+        retrieve_irf(dataset_model, dataset, dataset_model.get_global_dimension())
```
jsnel, line 88: Why is this at a different indentation level as the rest of the code inside the `if not is_full_model:` block. 
It appears as if the code was written in such a way that only if `not is_full_model` is true things are added to the result, so it's not clear what the intention is now. 

Please correct indentation or clarify in a comment.

glotaran/builtin/megacomplexes/decay/util.py
```diff
@@ -326,10 +327,15 @@ def retrieve_irf(dataset_model: DatasetModel, dataset: xr.Dataset, global_dimens
...
         ).data,
     )
+
     center = irf.center if isinstance(irf.center, list) else [irf.center]
     width = irf.width if isinstance(irf.width, list) else [irf.width]
     dataset["irf_center"] = ("irf_nr", center) if len(center) > 1 else center[0]
     dataset["irf_width"] = ("irf_nr", width) if len(width) > 1 else width[0]
+
+    if irf.shift is not None:
+        dataset["irf_shift"] = (global_dimension, [center[0] - p.value for p in irf.shift])
```
jsnel, line 337: It's hard (especially without unit test) to predict the types here, could you indicate in a comment the structure and types you expect so that when typing is added or unit tests are created this can be taken into account?

jsnel (changes requested):
Some minor clarification requested, otherwise it is ok code wise. 

I do note that unit tests are missing, and no integration tests (pyglotaran-examples) exists for this functionality. However since the code here appears necessary to check of an example is working correctly, it can be accepted if unit tests based on that (new) example are then created in a future PR.

glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
s-weigand: To me, it looks like that if an IRF is present its data will always be added to the result no matter if it is a full model or not (if not already added before).

glotaran/builtin/megacomplexes/decay/util.py
s-weigand: Just looked it up `irf.center` is always a `list[Parameter]` and `shift` is `list[Parameter] | None`

```python

@model_item(
    properties={
        "center": List[Parameter],
        ...
        "shift": {"type": List[Parameter], "allow_none": True},
        ...
    },
    has_type=True,
)
class IrfMultiGaussian:
```

s-weigand (changes requested):
Since we already use shift in one of our "unit" tests
https://github.com/glotaran/pyglotaran/blob/697df869288d2fe9fb80a1cff2ed0ee2aeaa8196/glotaran/builtin/megacomplexes/decay/test/test_decay_megacomplex.py#L90
it should be straightforward to also add a test for this feature inside the existing test.

glotaran/builtin/megacomplexes/coherent_artifact/coherent_artifact_megacomplex.py
joernweissenborn: This is simply that I removed the check for if the irf has already been done to the retrieve irf util so the coherent artifact can import and use it without me geting a banghammer for duplicatioin

s-weigand (approved):
I added a basic unittest, so from my side, it is good for merge 👍

But maybe `irf_shifted_center`/`shifted_irf_centers` would be a more fitting name... naming things 🤷‍♀️

jsnel:
> I added a basic unittest, so from my side, it is good for merge 👍
> 
> But maybe `irf_shifted_center`/`shifted_irf_centers` would be a more fitting name... naming things 🤷‍♀️

It represents a (per wavelength) shift that's added to the irf center, so it is aptly named. But good you mention it so I checked.

jsnel (approved):

PR #1022, opened 2022-03-25
👌🧹 Add suffix to rate and lifetime and guard for missing datasets

### LINKED ISSUE #1021: 🐛 Lifetimes are only saved for first megacomplex

joernweissenborn:
### Change summary

- lifetime and rate coordinate now get suffixes 
- suffixes are now always added for megacomplex specific data like DAS
- initial concentration has been added to result data
- a guard for missing data has been added

### Checklist

### Closes issues

closes #1021

### DIFF
--- a/.github/test_result_consistency.py
+++ b/.github/test_result_consistency.py
@@ -9,6 +9,7 @@ from functools import lru_cache
 from pathlib import Path
 from textwrap import dedent
 from typing import TYPE_CHECKING
+from typing import Iterable
 from typing import Protocol
 from warnings import warn
 
@@ -120,6 +121,30 @@ def get_current_result_path() -> Path:
         raise ValueError(f"No current results present, {RUN_EXAMPLES_MSG}")
 
 
+def rename_with_suffix(
+    expected_name: str, suffixed_names: Iterable[str], current_keys: Iterable[str]
+) -> str:
+    """Prepace ``expected_name`` with the suffixed version in the dataset keys.
+
+    Parameters
+    ----------
+    expected_name: str
+        Expected name of a variable (data_var, coord)
+    suffixed_names: list[str]
+        Names that are allowed/expected to have suffixes.
+    current_keys: Iterable[str]
+        Keys of the current dataset.
+
+    Returns
+    -------
+    str
+        Updated expected var name.
+    """
+    if expected_name in suffixed_names:
+        return next((key for key in current_keys if key.startswith(expected_name)), expected_name)
+    return expected_name
+
+
 def coord_test(
     expected_coords: DataArrayCoordinates,
     current_coords: DataArrayCoordinates,
@@ -136,23 +161,47 @@ def coord_test(
         ):
             print(f"- allow missing coordinate: {expected_coord_name} in variable {data_var_name}")
             continue
+
+        multiplier = 1
+        if expected_coord_name in ["rate", "lifetime"]:
+            multiplier = -1
+
+        expected_coord_name = rename_with_suffix(
+            expected_coord_name,
+            [
+                "species",
+                "initial_concentration",
+                "component",
+                "rate",
+                "lifetime",
+                "to_species",
+                "from_species",
+            ],
+            current_coords.keys(),
+        )
+
         assert expected_coord_name in current_coords.keys(), (
             f"Missing coordinate: {expected_coord_name!r} in {file_name!r}, "
             f"data_var {data_var_name!r}"
         )
 
-        # assert (
-        #     expected_coord_value.dims == current_coords.dims
-        # ), f"Dimensions mismatch in {data_var_name!r}"
-
         if exact_match or expected_coord_value.data.dtype == object:
             assert np.array_equal(
-                expected_coord_value, current_coords[expected_coord_name]
-            ), f"Coordinate value mismatch in {file_name!r}, data_var {data_var_name!r}"
+                expected_coord_value, multiplier * current_coords[expected_coord_name]
+            ), (
+                f"Coordinate value mismatch in {file_name!r}, "
+                f"data_var {data_var_name!r} and {expected_coord_name=}"
+            )
         else:
             assert allclose(
-                expected_coord_value, current_coords[expected_coord_name], rtol=1e-5, print_fail=20
-            ), f"Coordinate value mismatch in {file_name!r}, data_var {data_var_name!r}"
+                expected_coord_value,
+                multiplier * current_coords[expected_coord_name],
+                rtol=1e-5,
+                print_fail=20,
+            ), (
+                f"Coordinate value mismatch in {file_name!r}, "
+                f"data_var {data_var_name!r} and {expected_coord_name=}"
+            )
 
 
 def data_var_test(
@@ -169,6 +218,18 @@ def data_var_test(
     # when weights are applied
     if expected_var_name == "weighted_data" and expected_var_name not in current_result.data_vars:
         return
+
+    expected_var_name = rename_with_suffix(
+        expected_var_name,
+        [
+            "decay_associated_spectra",
+            "decay_associated_images",
+            "a_matrix",
+            "k_matrix",
+            "k_matrix_reduced",
+        ],
+        current_result.data_vars.keys(),
+    )
 
     assert (
         expected_var_name in current_result.data_vars
--- a/.github/test_result_consistency.py
+++ b/.github/test_result_consistency.py
@@ -162,10 +162,6 @@ def coord_test(
             print(f"- allow missing coordinate: {expected_coord_name} in variable {data_var_name}")
             continue
 
-        multiplier = 1
-        if expected_coord_name in ["rate", "lifetime"]:
-            multiplier = -1
-
         expected_coord_name = rename_with_suffix(
             expected_coord_name,
             [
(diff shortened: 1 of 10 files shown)


s-weigand (changes requested):
@joernweissenborn as discussed for consistency reasons please also add the suffix when only one megacomplex is used.

s-weigand:
Note from @joernweissenborn to himself lifetimes should be positive.

joernweissenborn:
@s-weigand all your changes have been included plus some more things, see description. So this is ready for final review.

requirements_dev.txt
```diff
@@ -10,7 +10,7 @@ numba==0.55.1
 numpy==1.21.5
 openpyxl==3.0.9
 pandas==1.4.1
-rich==11.2.0
+rich==12.0.1
```
jsnel, line 13: This is already part of main.

s-weigand: I might have rebased it onto the wrong branch 😖 
Resolved after rebasing on the [current main](https://github.com/glotaran/pyglotaran/commit/c8b993b753a4fe9cba64a1c5bd0fb461f3dac05c).

s-weigand (approved):
LGTM 👍

@jsnel An additional change is that:
- `rate_<megacomplex_name>`, `lifetime_<megacomplex_name>` now are positive

Coordinates with suffix:
- `species`
- `initial_concentration`
- `component`
- `rate`
- `lifetime`
- `to_species`
- `from_species`

Data variables with suffix:
- `decay_associated_spectra`
- `decay_associated_images`
- `a_matrix`
- `k_matrix`
- `k_matrix_reduced`

glotaran/builtin/megacomplexes/decay/util.py
```diff
@@ -271,42 +297,45 @@ def retrieve_decay_associated_data(
     matrix = k_matrix.full(species)
     matrix_reduced = k_matrix.reduced(species)
     a_matrix = megacomplex.get_a_matrix(dataset_model)
-    rates = k_matrix.rates(species, initial_concentration)
+    rates = -1 * k_matrix.rates(species, initial_concentration)
```
jsnel, line 300: This appears of the category, oh the answers is the wrong sign, so let's multiple it with another factor of -1 somewhere. But two wrongs don't make a right. I would expect a check somewhere to make sure the lifetime is positive when it should be, and that the rates that represent decays are of the appropriate sign when they should.

In any case why is it needed that rates is multiplied with -1, it should have been the correct sign already, and the multiplication should happen in the calculation of lifetimes. And even then it might make more sense to have a helper function that ensures correct sign, and throws a warning when something is wrong unexpectedly. 

In any case, all unit tests are passing so we have probably multiplied the correct number of times with -1, but I'd recommend a future refactor that solves this more elegantly and adds some unit tests.

jsnel (approved):
Overall ok, but left a comment suggesting some minor refactoring.

.github/test_result_consistency.py
```diff
@@ -136,23 +161,47 @@ def coord_test(
         ):
             print(f"- allow missing coordinate: {expected_coord_name} in variable {data_var_name}")
             continue
+
+        multiplier = 1
+        if expected_coord_name in ["rate", "lifetime"]:
+            multiplier = -1
```
s-weigand, line 167: @jsnel This is the part in the tests where we account for the changed sign of `rate` and `lifetime` (negative -> positive)

glotaran/builtin/megacomplexes/decay/util.py
jsnel: Actually, let us please create an issue motivating why the change is being made, and why before it was not as it should be, and now it is as it should be.

s-weigand (approved):
LGTM 👍

glotaran/builtin/megacomplexes/decay/util.py
ism200: By definition, the eigenvalues of the compartmental model are negative, and previously these eigenvalues were then reported as the rates. however, the rates are the negatives of the eigenvalues, thus this multiplication with -1 was overdue. then the lifetimes which are the reciprocals of the rates are also positive.  I recommend to add a comment line # the rates are the negatives of the eigenvalues of the compartmental model

s-weigand: @ism200 I think the point @jsnel wants to make is that `k_matrix.rates` should already return the proper rates instead of multiplying them with -1 right before we save them.
Current implementation:
https://github.com/glotaran/pyglotaran/blob/a635bb1dea965777fab77f70f536ae3c269c2e84/glotaran/builtin/megacomplexes/decay/k_matrix.py#L208-L219

ism200: in that case, you can put 
rates, _ = -self.eigen(compartments)
and add before line 218 a comment line # the rates are the negatives of the eigenvalues of the compartmental model

s-weigand: This should be fixed now ([ceed5cb](https://github.com/glotaran/pyglotaran/pull/1022/commits/ceed5cb34ca2faea0054c1cb820b9ff0c5792b47)-[40f6e75](https://github.com/glotaran/pyglotaran/pull/1022/commits/40f6e7526c55088a46f71b5d4f9be057f2727583))

s-weigand:
@joernweissenborn Since you are the lead developer plrease  review ([ceed5cb](https://github.com/glotaran/pyglotaran/pull/1022/commits/ceed5cb34ca2faea0054c1cb820b9ff0c5792b47)-[40f6e75](https://github.com/glotaran/pyglotaran/pull/1022/commits/40f6e7526c55088a46f71b5d4f9be057f2727583))

jsnel (dismissed):

glotaran/builtin/megacomplexes/decay/util.py
```diff
@@ -24,11 +24,10 @@ def calculate_matrix(
 ):
 
     compartments = megacomplex.get_compartments(dataset_model)
-    initial_concentration = megacomplex.get_initial_concentration(dataset_model)
     k_matrix = megacomplex.get_k_matrix()
 
     # the rates are the eigenvalues of the k matrix
-    rates = k_matrix.rates(compartments, initial_concentration)
+    rates, _ = k_matrix.eigen(compartments)
```
jsnel, line 30: in glotaran/builtin/megacomplexes/decay/k_matrix.py 
> rates are the negatives of the eigenvalues
that's not in line with the comment or the expression here.
@s-weigand can you double check?

s-weigand: True this could use some renaming... naming things ... `k_matrix_eigenvalues` 🙈
But since this is an internal and not a user-facing API we could maybe postpone this to a different issue?

ism200:
The minus sign is missing

Van: Sebastian Weigand ***@***.***>
Verzonden: Tuesday, March 29, 2022 12:21 PM
Aan: glotaran/pyglotaran ***@***.***>
CC: Stokkum, I.H.M. van ***@***.***>; Mention ***@***.***>
Onderwerp: Re: [glotaran/pyglotaran] 👌🧹 Add suffix to rate and lifetime and guard for missing datasets (PR #1022)


@s-weigand commented on this pull request.

________________________________

In glotaran/builtin/megacomplexes/decay/util.py<https://github.com/glotaran/pyglotaran/pull/1022#discussion_r837309743>:

>      k_matrix = megacomplex.get_k_matrix()







     # the rates are the eigenvalues of the k matrix



-    rates = k_matrix.rates(compartments, initial_concentration)



+    rates, _ = k_matrix.eigen(compartments)



True this could use some renaming... naming things ... k_matrix_eigenvalues 🙈
But since this is an internal and not a user-facing API we could maybe postpone this to a different issue?

s-weigand:
> The minus sign is missing 
@ism200 This is in the context of `calculate_matrix` where the term `rates` is misused and the eigenvalues are meant, which is why @jsnel flagged it.

ism200:
my understanding is that this is the most logical place to add the minus sign that is needed to compute the rates from the eigenvalues. then later the "rates" can directly be used to compute the lifetimes.

glotaran/builtin/megacomplexes/decay/util.py
joernweissenborn: nope, you must use the `rates` method. This ensures that in a sequential model what @ism200 calls the "B matrix". This has numerical advantages as it avoids calculating the eigenvalues.

s-weigand: @joernweissenborn How about we add an `eigenvalues` method that does the same as the old `rates` method and rates just returns `self.eigenvalues` ?

s-weigand: I discussed it with @joernweissenborn and dropped the commit in favor of improving the comment.

s-weigand:
I hope [9b12b5f](https://github.com/glotaran/pyglotaran/pull/1022/commits/9b12b5f0192b438eae0d41884f02de952bb3ac64) finally solved all the confusion about the rates 😅
Thnx @jsnel for insisting `There must be a better way`

jsnel (approved):
LGTM now, thanks for incorporating my requested changes:
-  👌 Removed sign confusion about rates (9b12b5f0192b438eae0d41884f02de952bb3ac64)

s-weigand (approved):

jsnel (approved):

PR #1047, opened 2022-04-03
♻️ Move optimization to new module `glotaran.optimization`

joernweissenborn:
### Change summary

- created new module optimization
- deprecated `glotaran.analysis.optimize`
- adapted all imports

### Checklist

### Closes issues

None.


s-weigand (approved):
In theory, we would need to deprecate all the moved modules not just `optimize`.
But since `optimize` contains the only user facing API and the rest is internal API only I think it would be overkill to deprecate them all.

LGTM 👍

jsnel (approved):

PR #1076, opened 2022-05-20
🩹 Show validation problem if parameters are missing values (default: NaN)

s-weigand:
This change fixes false positive reporting of a `Model` with parameters or `Scheme` as valid when parameters are missing values and default to `np.nan`.
One use case where a user would want missing parameters is in a teaching exercise where students have to fill in missing parameters.
The other use case would be that a user started editing the parameters file but didn't finish and forgot some values.

As a demo you can use [this notebook](https://github.com/ism200/PE2022/blob/3a5173cce2cf417f605a586c62beedb7a1af4e49/BRC-case-study/BRC_VIS-global-case-study-pretty.ipynb).

### Change summary

- [👌 Don't set default parameter value to zero](https://github.com/glotaran/pyglotaran/commit/e725f4b62e89dd54afd46cb7a92cdf36f2a0318e)
- [🧹 Improved typing and exception raising](https://github.com/glotaran/pyglotaran/commit/2e4416287c15b44f4368eb4e040bd46a3db7b438)
- [🩹 Fix degraded label with nesting over 2](https://github.com/glotaran/pyglotaran/commit/a051427ffd69ea54450618582ec2a96a17c2918f)
- [✨ Added missing_parameter_value_labels property to ParameterGroup](https://github.com/glotaran/pyglotaran/commit/aa17bf01a92aa35e4f04827e612fd302d93dd687)
- [✨ Added missing parameters check to model validation](https://github.com/glotaran/pyglotaran/commit/4a96db257fb49ca8eda7bff79fba335771bf66af)
- [👌 Improved validation output](https://github.com/glotaran/pyglotaran/commit/2acf3ecbc5f411ca9cabadf9de657f6cb13efcca)

### Checklist

### DIFF
--- a/glotaran/parameter/parameter_group.py
+++ b/glotaran/parameter/parameter_group.py
@@ -585,6 +585,21 @@ class ParameterGroup(dict):
                     )
                 parameter.value = value
 
+    @property
+    def missing_parameter_value_labels(self) -> list[str]:
+        """List of full labels where the value is a NaN.
+
+        This property is used to validate that all parameters have starting values.
+
+        Returns
+        -------
+        str
+            List full labels with missing value.
+        """
+        parameter_df = self.to_dataframe(as_optimized=False)
+        parameter_nan_value_mask = parameter_df["value"].isna()
+        return parameter_df[parameter_nan_value_mask]["label"].to_list()
+
     def markdown(self, float_format: str = ".3e") -> MarkdownStr:
         """Format the :class:`ParameterGroup` as markdown string.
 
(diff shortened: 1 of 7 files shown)


glotaran/parameter/parameter_group.py
```diff
@@ -132,10 +133,8 @@ def from_list(
 
         for i, item in enumerate(parameter_list):
             if isinstance(item, (str, int, float)):
-                try:
+                with contextlib.suppress(ValueError):
```
jsnel, line 136: This looks much better!

jsnel (approved):
Nice user experience improvement! 
Good tests and thanks for beautifying the code here and there.

PR #1152, opened 2022-10-28
🚇Add .venv folder (and derivatives) to .gitignore

jsnel:
This makes it possible to more easily work with python (venv) virtual environments.

Add `.venv*` to .gitignore which captures the [most common](https://docs.python.org/3/tutorial/venv.html) naming pattern for standard python virtual environments:
> A common directory location for a virtual environment is .venv. This name keeps the directory typically hidden in your shell and thus out of the way while giving it a name that explains why the directory exists. It also prevents clashing with .env environment variable definition files that some tooling supports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ pyvenv.cfg
 /venv*
 /.env/
 
+# sensitive environment variables
+.env
+
 # TIMP data
 *.RData
 *.Rhistory


s-weigand:
Guess we should also add `venv*` since `python -m venv venv` is also pretty common

jsnel:
> Guess we should also add `venv*` since `python -m venv venv` is also pretty common

Ok. I wanted to avoid adding something that could be a real filename, but I guess there's little risk we will have a file or folder named `venv_something_else.py`.

s-weigand:
As an inspiration [this is github default `.gitignore`](https://github.com/github/gitignore/blob/e5323759e387ba347a9d50f8b0ddd16502eb71d4/Python.gitignore#L122) for python repos

s-weigand:
> Ok. I wanted to avoid adding something that could be a real filename, but I guess there's little risk we will have a file or folder named `venv_something_else.py`.

You can always prefix `/` to signal that `git` should only look at the repo root to match the pattern.

jsnel:
> As an inspiration [this is github default `.gitignore`](https://github.com/github/gitignore/blob/e5323759e387ba347a9d50f8b0ddd16502eb71d4/Python.gitignore#L122) for python repos

I'd prefer not to use .env since that's also used by other things, and I'd like to keep the * in there because I like to use different names to manage different python versions, e.g.:
`.venv` (current version)
`.venv38`
`.venv310`
`.venv311`
`.venv-crazy-experiment`
... which this allows :)

s-weigand (approved):
LGTM 👍

Issue #749, opened 2021-07-15
💬 Add Issue Triage 


joernweissenborn:
We have tons of old issues, a bot ~~based on Dangerfile~~ could help.

s-weigand:
Needs more information on what we want and need to do with that.
If we just want to close stale issues we could use [Stale](https://github.com/marketplace/stale).
If we want to add labels based on the text there are also bots for that.
But we also use issues as a kind of backlog for future ideas... so it needs a clearer description.

s-weigand:
@joernweissenborn This is still missing information, feel free to close it if not relevant anymore

PR #1172, opened 2022-11-06
🧰 Stale bot

jsnel:
Automatically close stale Issues and Pull Requests that tend to accumulate during a project. 
As suggested here: https://github.com/glotaran/pyglotaran/issues/749#issuecomment-921319429 

> After a period of inactivity, a label will be applied to mark an issue as stale, and optionally post a comment to notify contributors that the Issue or Pull Request will be closed.
> 
> If the Issue or Pull Request is updated, or anyone comments, then the stale label is removed.
> 
> If no more activity occurs, the Issue or Pull Request will be automatically closed.
Ref: https://github.com/marketplace/stale

closes: #749

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 93
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 31
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - 'Type: Serious Bug'
+# Label to use when marking an issue as stale
+staleLabel: 'Status: Lack of interest'
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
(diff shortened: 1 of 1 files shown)


s-weigand:
Hmmm most of our issues are basically of the `Note to myself` nature since we don't have an active community opening issues YET.
That said I have seen maintainers spam issues with comments for the bot to reopen unresolved issues on rather big and active projects.
So I'm not too sure if we want and need it at this point in time.

s-weigand:
@jsnel Let's add this after the 0.7.0 release, and close as many "todo like" issues til then.

joernweissenborn:
So at my daytime job I wrote a python based helper for housekeeping in gitlab, also for issue triage. I have a very simple flow there:

- tag it if it has not been touched for 90 days
- deleted if it is not touched after that for 180 days

Result is basically that mostly I remove the label for those I want to keep in backlog to prevent removal and occasionally close an issue because it has been solved in other way or requirements  changed and what ever. 

So in the end it will mostly end in a kind of monthly check of issues with outdated label. Not a big deal as maintainer.

s-weigand (approved):
With the generous 3months until stale and 6 months after that until an issue is closed + filtering by labels I think we can merge this in.

PR #1174, opened 2022-11-10
👌 Allow more natural column names in pandas parameters file reading

s-weigand:
PR #1135 changed the column names from 'non-negative' -> 'non_negative' and 'standard-error' -> 'standard_error' which causes crashes in load_parameters when reading parameters files that were created before the merge (e.g. 'optimized_parameters.csv').

In addition to reenabling the reading of old parameter files saved with the `pandas` plugins (`csv`, `tsv`, `xlsx`, `ods`), this PR allows the same more natural/shorter column names which are allowed in `yaml` files for the `pandas` plugins and casts the column names to lower strings to counter capitalization problems caused by auto-capitalization from `excel` and the likes.

https://github.com/glotaran/pyglotaran/blob/e70f7603d8a55c82caec825938c9477f21d19aa1/glotaran/parameter/parameter.py#L39-L45

And it removed the unintended requirement to define the `non_negative` and `vary` columns, which are optional and can be filled with default values.

### Change summary

- [🩹 Fix reading of old pandas saved parameters files](https://github.com/glotaran/pyglotaran/pull/1174/commits/04f5d671f2376ca48d8917b318a0d2cf895215c6)
- [👌 Allow alternate notation used in yaml for pandas plugins](https://github.com/glotaran/pyglotaran/pull/1174/commits/cf33e1e58e3126c5c670310290af071ad835f608)
- [👌 Make column names case insensitive for pandas plugins](https://github.com/glotaran/pyglotaran/pull/1174/commits/cf33e1e58e3126c5c670310290af071ad835f608)
- [🧪 Added unittest for Parameters.from_dataframe column validation](https://github.com/glotaran/pyglotaran/pull/1174/commits/2af014358e413502b9be7d5a677b20f047037e10)

### Checklist


glotaran/parameter/parameters.py
```diff
@@ -157,6 +157,10 @@ def from_dataframe(cls, df: pd.DataFrame, source: str = "DataFrame") -> Paramete
 
         .. # noqa: D414
         """
+        df.rename(
```
joernweissenborn, line 160: I don't like `parameter` package be aware of this, this should be done within plugins.

joernweissenborn (changes requested):

jsnel (changes requested):
Agree, let's move the column renaming out of the parameters.py, validity of column names should be checked at parsing time (it's part of the user-facing API and should be validated, and errors (e.g. typos) reported back to the user. 
In this case, csv.py and xlsx.py can handle the renaming.

Some other remarks while reviewing related code:
- from_dataframe implicitly introduces required column names (e.g. minimum, maximum, non_negative, vary) which are not required in other input formats (e.g. yaml) because they have default values. This should not behave differently in that regard.
- please allow for shorthand notation as well (expr | expression, min | minimum, etc)

glotaran/parameter/parameters.py
jsnel: I included a comment about this in my review [here](https://github.com/glotaran/pyglotaran/pull/1174#pullrequestreview-1177935323).

jsnel:
Also please, if possible, check for (accidental) capitalization and provide feedback to the user (error in yaml, warning in csv/xlsx).  However, in the case of xlsx format just ignore it (lowercase it) (e.g. LibreOffice does auto capitalization of string labels).

glotaran/parameter/parameters.py
s-weigand: Fixed in [6aa60e6](https://github.com/glotaran/pyglotaran/pull/1174/commits/6aa60e6b6f51b1d8b6b7a7d03364d58c2fc6dffb)

jsnel (approved):

PR #1229, opened 2023-02-04
✨ Integrate plugin system into Project

s-weigand:
This change allows the project API to leverage the full plugin system to detect and read
- Models
- Parameters
- Data

### Change summary

- [♻️ Changed plugin_system implementation check functions to generators](https://github.com/glotaran/pyglotaran/commit/810b40b1caae331888f546e1e4d16c912c1f9c6c)
- [♻️ Refactored test_base_registry and added test case for all methods implemented](https://github.com/glotaran/pyglotaran/commit/90e40f0069b218444281f26b92496d20398a3ed4)
- [✨ Added supported_file_extensions function to base_registry](https://github.com/glotaran/pyglotaran/commit/e622e6bbc57d1284488096736aee1f9299db1b91)
- [✨ Added supported_file_extensions function for project and data io](https://github.com/glotaran/pyglotaran/commit/92b98be70a1f69affd4bc42454f0de8e03be5c38)
- ♻️ Refactored plugin registry mocking using functions from glotaran.testing.plugin_system
- [✨ Integrated plugin system into Project](https://github.com/glotaran/pyglotaran/commit/7a4e3afbb1aa53faac7b27dbcd75f22984b51e4d)

### Checklist

### DIFF
--- a/glotaran/plugin_system/test/test_base_registry.py
+++ b/glotaran/plugin_system/test/test_base_registry.py
@@ -73,6 +73,7 @@ def get_mock_plugin_function(plugin_registry_key: str):
         return MockPluginSubclassFull
     elif plugin_registry_key == "sub_class_str":
         return MockPluginSubclassStr
+    raise ValueError(f"No mock plugin with name {plugin_registry_key!r}")
 
 
 mock_registry_data_io = cast(
--- a/glotaran/plugin_system/test/test_base_registry.py
+++ b/glotaran/plugin_system/test/test_base_registry.py
@@ -330,7 +330,7 @@ def test_methods_differ_from_baseclass(
     """Inherited methods are the same as base class and overwritten ones differ"""
     result = methods_differ_from_baseclass(method_names, plugin, MockPlugin)
 
-    assert result == expected
+    assert list(result) == expected
 
 
 @pytest.mark.parametrize(
@@ -367,7 +367,7 @@ def test_methods_differ_from_baseclass_table(
         method_names, plugin_registry_keys, get_plugin_function, MockPlugin
     )
 
-    assert result == expected
+    assert list(result) == expected
 
 
 def test_methods_differ_from_baseclass_table_plugin_names():
@@ -383,4 +383,4 @@ def test_methods_differ_from_baseclass_table_plugin_names():
         "some_method", "base", get_plugin_function, MockPlugin, plugin_names=True
     )
 
-    assert result == [["`base`", False, "`test_base_registry.MockPlugin`"]]
+    assert list(result) == [["`base`", False, "`test_base_registry.MockPlugin`"]]
--- a/glotaran/plugin_system/test/test_base_registry.py
+++ b/glotaran/plugin_system/test/test_base_registry.py
@@ -45,9 +45,26 @@ class MockPlugin:
         pass
 
 
-class MockPluginSubclass(MockPlugin):
+class MockPluginSubclassPartial(MockPlugin):
     def some_method(self):
         return "different implementation"
+
+
+class MockPluginSubclassFull(MockPlugin):
+    def some_method(self):
+        return "different implementation"
+
+    def some_other_method(self):
+        return "different implementation"
+
+
+def get_mock_plugin_function(plugin_registry_key: str):
+    if plugin_registry_key == "base":
+        return MockPlugin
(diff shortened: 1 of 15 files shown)


glotaran/plugin_system/test/test_base_registry.py
```diff
@@ -45,11 +46,35 @@ def some_other_method(self):
...
+def get_mock_plugin_function(plugin_registry_key: str):
+    if plugin_registry_key == "base":
+        return MockPlugin
+    elif plugin_registry_key == "sub_class_partial":
+        return MockPluginSubclassPartial
+    elif plugin_registry_key == "sub_class_full":
+        return MockPluginSubclassFull
+    elif plugin_registry_key == "sub_class_str":
+        return MockPluginSubclassStr
+
```
jsnel, line 76: I know it's just a mock function used in testing but maybe fail with not implemented error when called with unsupported key?

glotaran/plugin_system/test/test_base_registry.py
```diff
@@ -330,22 +357,23 @@ def test_methods_differ_from_baseclass(
     """Inherited methods are the same as base class and overwritten ones differ"""
     result = methods_differ_from_baseclass(method_names, plugin, MockPlugin)
 
-    assert result == expected
+    assert list(result) == expected
```
jsnel, line 360: Why not change expected to also be a list?

glotaran/plugin_system/test/test_data_io_registration.py
```diff
@@ -25,17 +25,19 @@
...
     from typing import Any
 
     from _pytest.capture import CaptureFixture
-    from _pytest.monkeypatch import MonkeyPatch
 
     from glotaran.typing import StrOrPath
 
 
-class MockDataIO(DataIoInterface):
+class MockDataIo(DataIoInterface):
```
jsnel, line 40: I understand you have adopted the "io" is a word hence the capitalization is Io instead of IO, but widely adopted convention states the following: 

> When using acronyms, use Pascal case or camel case for acronyms more than two characters long. For example, use HtmlButton or htmlButton. However, you should capitalize acronyms that consist of only two characters, such as System.IO instead of System.Io.
Ref: https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN

See also discussion here: https://stackoverflow.com/a/15526526

glotaran/testing/plugin_system.py
```diff
@@ -49,10 +49,10 @@ def _monkeypatch_plugin_registry(
     monkeypatch_plugin_registry_project_io
     """
     if test_registry is not None:
-        initila_plugins = {} if create_new_registry else __PluginRegistry.__dict__[register_name]
+        initial_plugins = {} if create_new_registry else __PluginRegistry.__dict__[register_name]
```
jsnel, line 52: whoops, funny typo, good catch ^^

jsnel (changes requested):
Looks, ok, please review for small changes.
Don't worry about the Io->IO, that's just a remark for now.

glotaran/testing/plugin_system.py
s-weigand: Wouldn't have seen it w/o spell checker 😅

glotaran/plugin_system/test/test_data_io_registration.py
s-weigand: Not sure if this applies here since the `Io` is not stand alone. 🤔
Also, Python naming conventions aren't the same as `dotnet`.
In any case, this change makes it consistent with `DataIoInterface`, where a change to `DataIOInterface` would make it harder to read.

glotaran/plugin_system/test/test_base_registry.py
s-weigand: `expected` is a list, but `result` is a generator so we need to create a list from it to execute it.

glotaran/plugin_system/test/test_base_registry.py
s-weigand: Fair enough fixed it in [d03982f](https://github.com/glotaran/pyglotaran/pull/1229/commits/d03982f090549ae82893eee5965249c7c27a32ac).

jsnel (approved):
LGTM!

PR #1256, opened 2023-02-23
Feature: Add dataprocessor.

joernweissenborn:
Adds a (chainable) PreProcessingPipeline with (initial) support for CorrectBaselineAverage and CorrectBaselineValue.

### DIFF
--- a/glotaran/io/preprocessor/__init__.py
+++ b/glotaran/io/preprocessor/__init__.py
@@ -1,0 +1,2 @@
+"""Tools for data pre-processing."""
+from glotaran.io.preprocessor.pipeline import PreProcessingPipeline
--- a/glotaran/io/preprocessor/pipeline.py
+++ b/glotaran/io/preprocessor/pipeline.py
@@ -66,14 +66,14 @@ class PreProcessingPipeline(BaseModel):
 
     def correct_baseline_average(
         self,
-        selection: dict[str, slice | list[int] | int] | None = None,
+        select: dict[str, slice | list[int] | int] | None = None,
         exclude: dict[str, slice | list[int] | int] | None = None,
     ) -> PreProcessingPipeline:
         """Correct a dataset by subtracting the average over a part of the data.
 
         Parameters
         ----------
-        selection: dict[str, slice | list[int] | int] | None
+        select: dict[str, slice | list[int] | int] | None
             The selection to average as dictionary of dimension and indexer.
             The indexer can be a slice, a list or an integer value.
         exclude: dict[str, slice | list[int] | int] | None
@@ -84,5 +84,5 @@ class PreProcessingPipeline(BaseModel):
         -------
         PreProcessingPipeline
         """
-        self._push_action(CorrectBaselineAverage(exclude=exclude, selection=selection))
+        self._push_action(CorrectBaselineAverage(exclude=exclude, select=select))
         return self
--- a/glotaran/io/preprocessor/pipeline.py
+++ b/glotaran/io/preprocessor/pipeline.py
@@ -65,19 +65,24 @@ class PreProcessingPipeline(BaseModel):
         return self
 
     def correct_baseline_average(
-        self, selection: dict[str, slice | list[int] | int]
+        self,
+        selection: dict[str, slice | list[int] | int] | None = None,
+        exclude: dict[str, slice | list[int] | int] | None = None,
     ) -> PreProcessingPipeline:
         """Correct a dataset by subtracting the average over a part of the data.
 
         Parameters
         ----------
-        selection: dict[str, slice | list[int] | int]
+        selection: dict[str, slice | list[int] | int] | None
             The selection to average as dictionary of dimension and indexer.
+            The indexer can be a slice, a list or an integer value.
+        exclude: dict[str, slice | list[int] | int] | None
+            Excluded regions from the average as dictionary of dimension and indexer.
             The indexer can be a slice, a list or an integer value.
 
         Returns
         -------
         PreProcessingPipeline
         """
-        self._push_action(CorrectBaselineAverage(selection=selection))
+        self._push_action(CorrectBaselineAverage(exclude=exclude, selection=selection))
         return self
--- a/glotaran/io/preprocessor/pipeline.py
+++ b/glotaran/io/preprocessor/pipeline.py
@@ -1,0 +1,83 @@
+"""A pre-processor pipeline for data."""
+from __future__ import annotations
+
+from typing import Annotated
+
+import xarray as xr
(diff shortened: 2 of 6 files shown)


glotaran/io/preprocessor/pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Returns
+        -------
+        PreProcessingPipeline
+        """
+        self._push_action(CorrectBaselineValue(value=value))
+        return self
+
+    def correct_baseline_average(self, interval: dict[str, slice]) -> PreProcessingPipeline:
+        """Correct a dataset by subtracting an average of the data.
```
jsnel, line 68:
```suggestion
        """Correct a dataset by subtracting the average over a part of the data.
```

glotaran/io/preprocessor/pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Parameters
+        ----------
+        interval: dict[str, slice]
+            The intervals to average.
```
jsnel, line 73:
```suggestion
            The intervals to calculate the average over to subtract from the data.
```

glotaran/io/preprocessor/preprocessor.py
```diff
@@ -0,0 +1,78 @@
...
+            The data to process.
+
+        Returns
+        -------
+        xr.DataArray
+        """
+        selection = data
+        for axis, interval in self.interval.items():
+            selection = selection.sel({axis: interval})
+        return data - (selection.sum() / selection.size)
```
jsnel, line 78: This code needs some revision, it suggests that it works with more than one intervals, but it doesn't.

Suppose an time axis [0,100], are you intending it to work, [0,10] and [90,100] in one operations, or only [0,10] and then [90,100] in two operations. It's not clear, and there's no tests for this either.

jsnel (changes requested):
Nice start, but I'd like to see a bit more meaning full tests. 
The CorrectBaselineValue method is ok, just needs more tests.
The CorrectBaseLineAverage case needs revision, and more extensive tests.

glotaran/io/preprocessor/preprocessor.py
```diff
@@ -0,0 +1,77 @@
...
+        ----------
+        data: xr.DataArray
+            The data to process.
+
+        Returns
+        -------
+        xr.DataArray
+        """
+        selection = data.sel(self.selection or {}).drop_sel(self.exclude or {})
+        return data - (selection.sum() / selection.size)
```
s-weigand, line 77:
```suggestion
        return data - selection.mean()
```

glotaran/io/preprocessor/preprocessor.py
joernweissenborn: I also added `exclude`, if you eg want to average only the edges.

glotaran/io/preprocessor/test/test_preprocessor.py
```diff
@@ -0,0 +1,44 @@
...
+def test_correct_baseline_value():
+    pl = PreProcessingPipeline()
+    pl.correct_baseline_value(1)
+    data = xr.DataArray([[1]])
+    result = pl.apply(data)
+    assert result == data - 1
+
+
+@pytest.mark.parametrize("indexer", (slice(0, 2), [0, 1]))
+def test_correct_baseline_average(indexer: slice | list[int]):
```
s-weigand, line 16: Please also add a test for exclude and one that shows a combination of selection and exclude.

glotaran/io/preprocessor/preprocessor.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        return data - self.value
+
+
+class CorrectBaselineAverage(PreProcessor):
+    """Corrects a dataset by subtracting the average over a part of the data."""
+
+    action: Literal["baseline-average"] = "baseline-average"
+    selection: dict[str, slice | list[int] | int] | None = None
+    exclude: dict[str, slice | list[int] | int] | None = None
```
s-weigand, line 62: For consistency, we should name the arguments
- select
- exclude
Or
- selection
- exclusion

s-weigand (changes requested):
Looks like a good start, just got my usual nitpicks.

I guess for now there be no other way to compose pipelines than copypasting a file together or combining the action attributes.

```py
compose_pipeline = PreProcessingPipeline(pl1.actions[:-1] + pl2.actions[-2:])
```

glotaran/io/preprocessor/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Tools for data pre-processing."""
+from glotaran.io.preprocessor.pipeline import PreProcessingPipeline
```
s-weigand, line 2: `CorrectBaselineValue` and `CorrectBaselineAverage` also be reexported from here for ease of use.

s-weigand (changes requested):
Sound something else (sorry for not catching it before 😢)

glotaran/io/preprocessor/__init__.py
joernweissenborn: Nobody should use them directly, it makes no sense. They are considered internal.

s-weigand: Yeah nvm I'm just tired

s-weigand (approved):
Looks good to me now, thanks for adding the preprocessing. 🎉

jsnel (dismissed):
It may not be the most intuitive way of doing preprocessing, but it works (for the limited cases currently defined), and it doesn't conflict with anything, so I'm ok with merging it in. 

I do think improvements are possible, but that can be tackled in a future PR.

glotaran/io/preprocessor/preprocessor.py
jsnel: The code is more correct and clear now than before.
```python
return data - data.sel(self.select or {}).drop_sel(self.exclude or {}).mean()
```

glotaran/io/preprocessor/pipeline.py
jsnel: The suggestions were incorportated.

glotaran/io/preprocessor/pipeline.py
jsnel: The suggestions were incorporated.

glotaran/io/preprocessor/test/test_preprocessor.py
jsnel: We'll get back to it in a future PR ;)

jsnel (approved):
Re-approve changes following dropped commit.

PR #1276, opened 2023-04-10
🧪 Test numba 0.57.0 result comparison

s-weigand:
Since `numba` is one of our core dependencies and currently the blocker to support python 3.11, we should help them test their release candidate.

### Checklist


s-weigand:
Just trying out numba 0.57.0 on py311 and comparing results

s-weigand:
This PR served its purpose by showing that we get the same results on python 3.11 with numba 0.57.0 and we did before.

PR #1379, opened 2023-10-21
🧰 Use ruff as linter

s-weigand:
This replaces flake8 in favor of [ruff](https://docs.astral.sh/ruff/).

Not all the rules we want are activated yet, but it is a start and more thorough than what flake8 checked before.

When reviewing have a look at the [rule sets which are applied](https://docs.astral.sh/ruff/rules/).

Please take special note of the comment in https://github.com/glotaran/pyglotaran/commit/29885ccb03c6b8e39e3c89711633dd08f5ad59b9


joernweissenborn (approved):
Nice PR, thx a lot!

joernweissenborn (approved):

jsnel (approved):
Good stuff! Love ruff!

PR #1381, opened 2023-10-21
♻️ Change Parameter from attrs to pydantic

s-weigand:
No need to have `attrs` and `pydantic` at the same time.

PR #1382, opened 2023-10-21
🩹 Fix schema generation by making private attributes private

s-weigand:
With pydantic 2 we had crashes when calling `.model_json_schema()` e.g.

```py
from glotaran.project import Scheme

Scheme.model_json_schema()
```

The main reason for this was pydatic failure to serialize `data_model_type` to json since the value is a class.

~~This can all be prevented by making internal only fields "private" (leading `_`) so they get excluded from the schema since we don't want users to put values for them anyway.~~

As @joernweissenborn  noted properly typing the fields with `ClassVar` those variables will also not be considered for the schema.

### DIFF
--- a/glotaran/builtin/elements/baseline/element.py
+++ b/glotaran/builtin/elements/baseline/element.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 class BaselineElement(Element):
     type: Literal["baseline"]  # type:ignore[assignment]
-    _register_as: ClassVar[str] = "baseline"
+    register_as: ClassVar[str] = "baseline"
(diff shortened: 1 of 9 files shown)


s-weigand:
This PR is a draft since it depends on #1381

jsnel:
> This PR is a draft since it depends on #1381

That PR has been merged, so promoting this PR.

jsnel (approved):
LGTM. Examples ran successfully.

joernweissenborn (approved):
Lgtm, but in the long run we should put these things into a `ElementConfig` class.

PR #1384, opened 2023-10-22
🩹 Fix dataset weights

jsnel:
Tested with freshly ported ex_two_datasets example


s-weigand (approved):
Seams right since `get_model_weight(model)` creates a new array without having side effects. Good catch 👍

joernweissenborn (approved):
Thx for fixing.

PR #1385, opened 2023-10-22
✨ Create useful schema

s-weigand:
This PR adds the `create_model_scheme_json_schema` function in `glotaran.utils.json_schema` to create json schema for the scheme file.
When `create_model_scheme_json_schema` is used with `Parameters` the parameter labels will autocomplete in the editor where parameters can be used.

Example usage in the staging repo (run in repo root):
```py
from glotaran.utils.json_schema import create_model_scheme_json_schema

create_model_scheme_json_schema(
    "GFP-for-rewrite/models/schema.json",
    "GFP-for-rewrite/models/parameters_sequential2.yml",
)
```

Add the following at the top of the converted schema files (models):
```yaml
# yaml-language-server: $schema=schema.json
``` 

There still are some refinements to do but this is at least a start.


glotaran/utils/json_schema.py
```diff
@@ -0,0 +1,121 @@
...
+    json_schema["$defs"] |= data_model_schema.pop("$defs")
+    json_schema["$defs"]["GlotaranDataModel"] = data_model_schema
+    json_schema["$defs"]["ExperimentModel"]["properties"]["datasets"]["additionalProperties"][
+        "$ref"
+    ] = "#/$defs/GlotaranDataModel"
+
+    # We need this overwrite since we generate json schema for editor support (auto completion and
+    # linting) of model schema files where the parameter label is used rather than defining a
+    # parameter.
+    parameter_label_schema: dict[str, str | list[str]] = {
```
joernweissenborn, line 103: what about constant parameters? The schema should indicate to the user that she can put just a numeric constant as well.

joernweissenborn, line 108: Another question: why not make this parameter schema things part of the `ParameterType` alias?

s-weigand: That is a different schema route, this just overwrites what "Parameter" means.
E.g. this is the schema of `rates` inside of the `KineticElement` schema, where `Parameter` is only one option in a union type.
```json
"rates": {
    "additionalProperties": {
        "anyOf": [
            {
                "$ref": "#/$defs/Parameter"
            },
            {
                "type": "number"
            },
            {
                "type": "string"
            }
        ]
    },
    "title": "Rates",
    "type": "object"
},
```

jsnel (approved):
I saw it in action. Nice improvement, especially for VS Code users.

joernweissenborn (approved):
I still think it would be more elegant to facilitate the `pydantic` hook to modify the result of calling `x.schema()`, but I can live with that.

UPDATE: I know this is made for creating schemas for projects, but I still like a more general schema for `Scheme` etc which can be even uploaded and recognized in the future.

s-weigand:
The schema is valid, just ill-suited for the editor implementation, but you could also generate schema for a different purpose (e.g. web API) and initializing from a parameter dict works perfectly fine which is why I decided to manipulate the schema dict instead of changing the `json_schema_extra` for `Parameter` itself
```py
from glotaran.builtin.elements.kinetic import KineticElement

KineticElement(
        **{
            "label": "sequential",
            "type": "kinetic",
            "rates": {
                ("s2", "s1"): {"label":"rates.1","value":5},
            },
        }
    )
```
This will return the following
```py
KineticElement(rates={('s2', 's1'): Parameter(label='rates.1', value=5.0)}, type='kinetic', dimension='time', label='sequential', extends=None)
```

PR #1427, opened 2024-02-02
🧹 Use element plugin registry to define LibraryType

s-weigand:
By using the plugin registry instead of looking up inheritance we can ensure that only elements that were explicitly added are valid types (e.g. `ExtendableElement` is not instantiable by itself and should not be part of `LibraryType`).

### Change summary

- [🧹 Use element plugin registry to define LibraryType](https://github.com/glotaran/pyglotaran/commit/becc4b6a88232169950f32a05674eb769b967182)
- [🧹 Remove InternalMockElement](https://github.com/glotaran/pyglotaran/commit/fa8b6775b939f12dc03bf5b4d44e9afc4f8968ad)
- [🧰⬆️ Update pre-commit config](https://github.com/glotaran/pyglotaran/commit/28e405141889baac9b0378b3ffb97838223a707a)

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
         args: [--remove]
 
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: "1.5.1"
+    rev: "1.7.0"
     hooks:
       - id: pyproject-fmt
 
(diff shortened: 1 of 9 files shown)


joernweissenborn (approved):
LGTM,Was just a quick and dirty placeholder anyway.

PR #1500, opened 2024-07-13
🩹Fix order of parameters involved in parameters expresssion

### LINKED ISSUE #1499: [staging] 🐛 Parameters expressions are not evaluated in the correct order
### Version information

- pyglotaran version: staging (v0.8.0-dev)

### Describe the bug

We observed there were differences in the optimization between main (0.7) and staging (0.8-dev) and found the problem to be related to models in which parameters relations (expression) were used. 

After some debugging we found that the order of execution (evaluating parameters) was incorrect, the parameter that had dependencies would itself be added *before* its dependencies causing it to be evaluated with values from a previous iteration. 

### To Reproduce

Run any model that has parameters relations in it, for instance:
`study_transient_absorption/transient_absorption_target_analysis.ipynb
on main: [transient_absorption_target_analysis.ipynb`](https://github.com/glotaran/pyglotaran-examples/blob/main/pyglotaran_examples/study_transient_absorption/transient_absorption_target_analysis.ipynb)
on staging: [transient_absorption_target_analysis](https://github.com/glotaran/pyglotaran-examples/blob/staging_rewrite/pyglotaran_examples/study_transient_absorption/transient_absorption_target_analysis.ipynb)

Then observe that the optimization (given sufficient number of iterations) is following a different trajectory.
If no update to the relation parameter happens (e.g. it is fixed, or there are not sufficient iterations/function evaluations possible) you will not see it. Also the first iteration is always idential because the initial parameter values are used.

Important note is that on main the order in which parameters are evlauated is the order in which they were put in the parameters specification file, thus to ensure correctness, relation parameters need to be before the parameters in which they are used in an expresssion. 

### Expected behavior

Both main and staging follow the same optimization trajectory and arrive at the same final result.

jsnel:
Add the dependent parameters after its dependencies.

Parameters involved in a relation (parameter expression) should be added before the parameter that declares a dependency on them.

### Change summary

- 🩹Fix order of parameters involved in parameters expresssion

### Checklist

### Closes issues

closes #1499

### DIFF
--- a/tests/model/test_item.py
+++ b/tests/model/test_item.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import random
+from itertools import permutations
 from typing import Literal
 
 from glotaran.model.item import Item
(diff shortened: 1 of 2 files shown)


jsnel:
Status of the compare results test

Currently on staging:
================ 159 failed, 418 passed, 49 warnings in 14.00s =================

With the changes in this PR: 
================ 148 failed, 429 passed, 49 warnings in 13.75s =================

jsnel:
TODO: add some unit tests for parameter resolving (this is why PR is still in DRAFT)

jsnel:
Reduced the number of errors in the comparison further by [adjusting the order of the scheme definition](https://github.com/glotaran/pyglotaran-examples/commit/307d8cf27bd3df98873e48ae5959453f78fb9333) in the fluorescence example to preserve the order of compartments.

New status:
================ 141 failed, 436 passed, 49 warnings in 14.14s =================

joernweissenborn:
This is actually lingering much longer, like very long. I never thought much about it but yeah, parameter order matters with related stuff.

The fix is totally correct. Would lgtm it now but you want to add a test.

jsnel:
Unit tests were added to check (specifically) for the issue that this PR fixes (incorrect order of parameters). 

The final result remains unchanged:
================ 141 failed, 436 passed, 49 warnings in 13.83s =================

The PR is ready for review now.

joernweissenborn (approved):
Needs to have a follow up pr to make the underlying system more robust

tests/model/test_item.py
```diff
@@ -136,3 +138,105 @@ def test_resolve_item_parameters():
...
+    ]
+}
+
+
+def _add_parameters_to_initial(labels, parameters, initial_parameters):
+    for label in labels:
+        add_to_initial(label, parameters, initial_parameters)
+
+
+def _test_add_to_initial(initial_parameters):
```
s-weigand, line 165: I get that we call `check_parameter_order` multiple times but just defining `expected_order` and using it multiple times should be fine when you also inline the key retrival.
As for the other functions I see no need at all to define functions in a closure vs. just inlining it since they are only called once.

tests/model/test_item.py
```diff
@@ -136,3 +138,105 @@ def test_resolve_item_parameters():
...
+        ), f"Incorrect expression for rates.k2. Got: {parameters.get('rates.k2').expression}"
+
+    test_ordered_addition()
+    test_unordered_addition()
+    test_partial_addition()
+    parameters = test_randomized_addition()
+    check_expressions(parameters)
+
+
+def test_add_to_initial_b_first():
```
s-weigand, line 231: Can you please add docstrings to the tests that explain the difference (in order vs. out of order definition)

s-weigand (changes requested):
Just some minor nitpicks about the tests. Sorry for that 😅

tests/model/test_item.py
jsnel: Addressed in [fe69696](https://github.com/glotaran/pyglotaran/pull/1500/commits/fe69696ed3be26218c0d80c756c62dbcc3045f81)

tests/model/test_item.py
jsnel: Addressed in [fe69696](https://github.com/glotaran/pyglotaran/pull/1500/commits/fe69696ed3be26218c0d80c756c62dbcc3045f81) where the tests are now inlined. 

The expected_order is not always the same, so that was kept, to be explicit about the expected order per test.

s-weigand (approved):
LGTM 👍

PR #1511, opened 2024-08-17
🧹🚇 Remove asv benchmarks and binder integration

s-weigand:
Since the 0.8 rewrite will drop ASV benchmarks there is no point in sticking to them and maintaining them on `main`.
Binder has not been functional for quite some time (startup timeout), so there is no point in keeping it.

### Change summary

- [🧹 Remove ASV benchmarks](https://github.com/glotaran/pyglotaran/commit/e6417229c90d49c70129404ad29e83c348b0d600)
- [🧹 Remove binder integration](https://github.com/glotaran/pyglotaran/commit/8c916ab6ffc40d3949ee76f839a7dd6391660532)

### Checklist


jsnel (approved):
👍 

In due time we'll look into establishing new benchmarks.

PR #1510, opened 2024-08-16
✨ Add support for pfid megacomplex

jsnel:
Add support for the modelling of perturbed free induction decay (PFID) as described in the 1995 paper by P. Hamm.

> P. Hamm, “Coherent effects in femtosecond infrared spectroscopy,” Chemical Physics 200, 415–429 (1995).
> DOI: [10.1016/0301-0104(95)00262-6](https://doi.org/10.1016/0301-0104(95)00262-6)

### Change summary

- Added PFID megacomplex 

The feature will be documented as part of an upcoming publication.

### Checklist

- [x] ✔️ Passing the tests (mandatory for all PR's)
- [x] 🚧 Added changes to changelog (mandatory for all PR's)
- [x] 🧪 Adds new tests for the feature (mandatory for ✨ feature and 🩹 bug fix PR's)

PR #1513, opened 2024-08-23
🩹 Fix AttributeError validating bad DOAS user definition

s-weigand:
During the review of #1510 we found an inherited bug from the DOAS megacomplex where a faulty user definition would raise an `AttributeError` when the `ItemIssue` get converted to a string that is show to users.

### Change summary

- [🧪 Add failing test reproducing the AttributeError validating a bad DOAS user definition](https://github.com/glotaran/pyglotaran/commit/5409d4d163405d68c4b09036fc64d0643d3fc58f)
- [🩹 Fix AttributeError validating bad DOAS user definition](https://github.com/glotaran/pyglotaran/commit/1f697b98a2c703cc46ec04465f529980d3695e48)

### Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 ### 🩹 Bug fixes
 
 - 🩹 Fix ordering bug in MatrixProvider class (#1512)
+- 🩹 Fix AttributeError validating bad DOAS user definition (#1513)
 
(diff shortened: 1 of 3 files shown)


jsnel (approved):
Good catch, also fixed it in #1510

PR #1518, opened 2024-08-23
🧰 Fix pre-commit config issues

jsnel:
Fixes 2 issues with faulty pre-commit configuration
- error when initializing clean pre-commit environment due to yanking of types-all
- fail in interrogate pipeline on missing pkg_resources in some environments due to interrogate not declaring explicit dependency on setuptools


s-weigand (approved):
Rather than working around old bugs in `interrogate` I raised the version of the hook that has all recent bug fixes (no functional changes that matter for us, if at all), to prevent unneeded venv cache clutter.
Besides that thanks for the fix 🚀

PR #1520, opened 2024-08-25
🩹 Use svd function on raw numpy arrray instead of xarray dataarray

jsnel:
This should fix an issues with numpy > 2.0.1 and should be backwards compatible.

### Change summary

- In the prepare_dataset function where svd are added we use the .data attribute of the xarrray dataarray to get the numpy array to be used in the calculation. 

### Checklist

- [x] ✔️Test with both numpy 2.0.1 and numpy 1.26.4


jsnel:
I manually tested with both numpy 2.0.1 and numpy 1.26.4, running all examples notebooks, and observed no failing cell executions.

s-weigand (approved):
Thanks this will enable compatibility with `numpy>2` (as far as `numba` also supports it) 🎉

PR #1534, opened 2024-09-09
Refactor Element Result Saving

### LINKED ISSUE #1498: [staging] 🐛 Add back per element A-/K-Matrices to result dataset
Currently staging saves the combined A-/K-Matrices for all decay elements.
If we need those is something to be decided by @ism200 .

However, what we need for validations are data variables on a per-element basis with the following name pattern

- `a_matrix_<element_name>`
- `k_matrix_<element_name>`
- `k_matrix_reduced_<element_name>`

To get rid of the spammy `xarray` warning:

```py
UserWarning: Duplicate dimension names present: dimensions {'species'} appear more than once in dims=('species', 'species').
```

Please use `('to_species', 'from_species')` notation.

### `main`
image
### `staging`
image

joernweissenborn:
This PR refactors the way element results are stored. Now every element stores it's own items. They return

```
@dataclass
class ElementResult:
    amplitudes: dict[str, xr.DataArray]
    concentrations: dict[str, xr.DataArray]
    extra: dict[str, xr.DataArray] = field(default_factory=dict)
```

The Kinetic element eg, stores amplitudes and contractions each with key "species" and "decay". This means in the final result this will be stored as eg `species_associated_amplitude_ELEMENTNAME`. K matrices etc are part of extra and only get the element name suffixed.

IMPORTANT To avoid collisions, all coords which are not the model or the global dim also get suffixed with the name, `species` coord will be renamend to `species_ELEMENTNAME`.

### Checklist

### Closes issues

closes #1498


joernweissenborn:
I have checked the sonar cloud issues, this will be fixed in the cleanup when I add doc etc.

glotaran/optimization/objective.py
```diff
@@ -17,16 +16,32 @@
 from glotaran.parameter.parameter import Parameter
 
 if TYPE_CHECKING:
+    from glotaran.model.element import ElementResult
     from glotaran.model.experiment_model import ExperimentModel
     from glotaran.typing.types import ArrayLike
 
 
+def add_svd_to_result_dataset(dataset: xr.Dataset, global_dim: str, model_dim: str):
```
s-weigand, line 24: Please use `glotaran.io.prepare_dataset.add_svd_to_dataset` where the [issue that causes the CI to fail](https://github.com/glotaran/pyglotaran/actions/runs/10770808052/job/29864958225?pr=1534) was [already fixed on staging](https://github.com/glotaran/pyglotaran/commit/2ef4ceabb9a5d9f2bd8476a32dd1c10703dbf972).

joernweissenborn: I rather like to move the function from io to here to avoid dependency of optimization to anything other than model.

I will copy the code from io with the fixes of course. This was just a quick fix to get IO out of the equation for dev.

s-weigand: Please keep in mind that you can't just move the function completely (needs re import or deprecation), since [we use it down stream](https://github.com/glotaran/pyglotaran-extras/blob/ae84a0c50ed5d3368b36a95f2a0ee9f84e58cf74/pyglotaran_extras/plotting/plot_svd.py#L7).

jsnel:
After internal discussions we decided to pivot on how to save results, see my summarizing [comment](https://github.com/glotaran/pyglotaran/issues/1498#issuecomment-2351701265) on the issue.

joernweissenborn:
Also fixes #1537 now

PR #1591, opened 2025-12-30
🩹 Fix changed numpy repr breaks ascii saving

s-weigand:
This change is a bug fix for the `ascii` builtin writing plugin when using a recent `numpy` version.
Recently `numpy` changed it's string representation of array items from a python float string representation to the actual type the item pointer points to e.g. `3.14` -> `np.float64(3.14)`.
This is the correct type that can be round trip evaluated, but the `ascii` plugin (one of our oldest code pieces that we need to replace!) doesn't handle this change nicely, since it uses the `repr` (used for representation in e.g. tracebacks) rather than the string representation.

### Checklist

## Summary by Sourcery

Fix ASCII explicit file writing to use plain string representations compatible with recent NumPy changes and verify round-trip of coordinates in tests.

Bug Fixes:
- Ensure wavelength and time values in ASCII explicit files are written using standard string formatting instead of NumPy repr to remain compatible with newer NumPy versions.

Tests:
- Extend explicit file reader/writer test to assert that time and spectral coordinates are preserved on round-trip.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,8 @@
 
 ### 🩹 Bug fixes
 
+- 🩹 Fix changed numpy repr breaks ascii saving (#1591)
+
 ### 📚 Documentation
 
 ### 🗑️ Deprecations (due in 0.9.0)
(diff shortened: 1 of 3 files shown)


jsnel (approved):
Thanks for taking the time to fix this! 

We'll improve in the next major/minor release of pyglotaran, but for the next patch release, this is perfectly fine!

PR #1608, opened 2026-03-13
🩹 Fix array dimensionality issues in add_svd_to_dataset and simulation

s-weigand:
This PR fixes two dimension-order dependent bugs that showed up when data or simulated results were represented with a different axis order than the code assumed.

The first issue was in SVD preparation. `add_svd_to_dataset` used the raw underlying array, so datasets with transposed dimensions or custom left/right singular vector dimensions could result in an exception due to dimensionality mismatch.

```py
xarray.structure.alignment.AlignmentError: cannot reindex or align along dimension 'time' because of conflicting dimension sizes: {2, 3} (note: an index is found along that dimension with size=3)
```

The fix normalizes the input `DataArray` to the requested `lsv_dim`/`rsv_dim` order before calling `np.linalg.svd`.

The second issue was in the simulation noise generation. Noise was previously applied directly to the current array layout, which meant swapping `megacomplex` and `global_megacomplex` in an equivalent full model could produce different noisy results because the random draws were consumed in a different dimension order. The fix applies noise in a normalized dimension order and then transposes the result back, making seeded simulations deterministic regardless of whether the same model is expressed with swapped model/global megacomplex assignments.

### Before
image

### After
image

### Change summary

- Fix `add_svd_to_dataset` to compute SVD on data transposed to the requested left/right singular vector dimensions
- Add regression tests for `add_svd_to_dataset`, including transposed input, custom dims, and the no-recompute path
- Fix simulation noise application so seeded noise is independent of dataset dimension ordering
- Add a `FullModel`-based regression test showing that swapping `megacomplex` and `global_megacomplex` yields the same noisy simulation result

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SVD input orientation so singular vectors align with intended dimensions.
  * Fixed noise application to preserve correct axis alignment when adding noise.

* **Tests**
  * Added comprehensive tests for SVD behavior (default, transposed, custom names, skip-recompute).
  * Added test ensuring simulation consistency when swapping megacomplex configurations.

* **Documentation**
  * Added changelog entry and test module docstring.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
 ### 🩹 Bug fixes
 
 - 🩹 Add pandas 3 compatibility (#1607)
+- 🩹 Fix array dimensionality issues in add_svd_to_dataset and simulation (#1608)
 
 ### 📚 Documentation
 
--- a/glotaran/io/prepare_dataset.py
+++ b/glotaran/io/prepare_dataset.py
@@ -85,7 +85,9 @@ def add_svd_to_dataset(
     if data_array is None:
         data_array = dataset[name] if name != "data" else dataset.data
     if f"{name}_singular_values" not in dataset:
-        l, s, r = np.linalg.svd(data_array.data, full_matrices=False)
+        l, s, r = np.linalg.svd(
+            data_array.transpose(lsv_dim, rsv_dim).to_numpy(), full_matrices=False
+        )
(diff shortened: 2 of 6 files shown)


jsnel (approved):
Looks good now. The numpy.random.Generator comment is just a nit.

